PR #437, opened 2012-12-06
add window manager within middleware

overnin:
I tried to embed the window manager in a middleware in order to slow down my SMPP transport when the aggregator don't use SMPP's Throttling mechanism. It's a first try and questions are still open and i would need some directions on the 2 following points

1> How to cancel the following middlewareStack and the worker execution? Two options were identified returning None or throwing an exception like "StopPropagation". None is not very explicit which can lead to mistake when deploying pile of middlewares. Also there is this rule already in the code that 'Return value should never be none'. An exception "StopPropagation" is  more explicit but i don't get all the implication that it might have.

2> How to resume the middlewareStack execution when a message is due to be send? in my code i directly call the worker to send the message... but then i'm bypassing all the following middlewares of the worker.

### DIFF
--- a/vumi/middleware/base.py
+++ b/vumi/middleware/base.py
@@ -103,6 +103,10 @@ class BaseMiddleware(object):
         """
         return failure
 
+    def resume_handling(self, handle_name, message, endpoint):
+        return self.worker._middlewares.resume_handling(
+            self, handle_name, message, endpoint)
+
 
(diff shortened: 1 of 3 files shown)


hodgestar:
@overnin This branch still alive?

kyrelos:
@overnin is this branch still alive?

PR #558, opened 2013-05-29
Upgrade to latest raven client

vgeddes:
This will simplify a lot of the code in vumi/sentry.py and will be a better match for the latest version of our Sentry instance.


smn:
👀

PR #704, opened 2014-01-18
pypy support

jerith:
This probably needs a little more thought before we decide if we want to land it, but we can update this branch periodically for Travis pypy builds or whatever.


vumi/application/tests/test_sandbox.py
```diff
@@ -400,6 +400,11 @@ def test_sandbox_command_does_not_parse_timestamps(self):
 
 class JsSandboxTestMixin(object):
 
+    BIGGER_RLIMITS = {
+        "RLIMIT_STACK": [2 * 1024 * 1024] * 2,
+        "RLIMIT_AS": [256 * 1024 * 1024] * 2,
+    }
```
rudigiesler: What's the background of why we need bigger rlimits for these tests?

vumi/transports/apposit/tests/test_apposit.py
```diff
@@ -48,11 +49,6 @@ def setUp(self):
         self.transport_url = self.transport.get_transport_url()
         self.web_path = config['web_path']
 
-    @inlineCallbacks
-    def tearDown(self):
-        yield self.mock_server.stop()
-        yield super(TestAppositTransport, self).tearDown()
-
```
rudigiesler: There are quite a few cleanups like this. Are they necessary or is it just code cleanup?

vumi/transports/truteq/tests/test_truteq.py
```diff
@@ -280,8 +280,9 @@ def test_handle_inbound_sms(self):
             yield self.send(cmd)
             [warning] = logger.messages()
             self.assertEqual(
-                warning,
-                "Received unsupported message, dropping: %r." % (cmd,))
+                warning[:59],
+                "Received unsupported message, dropping: <MO command_id=103 ")
+        self.flushLoggedErrors()
```
rudigiesler: I'm not sure why this has changed?

rudigiesler:
Review done. Looks good, but just a few questions of things that I'm unsure of.

vumi/application/tests/test_sandbox.py
jerith: When we run these tests on pypy, node.js fails to start with the default (smaller) rlimits. My guess is that pypy has a larger memory footprint than CPython and the default rlimits are borderline for node.js+CPython already.

vumi/transports/apposit/tests/test_apposit.py
jerith: Mostly cleanup, but I made all these changes while I was trying to track down the open sockets we were leaking in the Riak manager tests.

vumi/transports/truteq/tests/test_truteq.py
jerith: The `repr()` of the command object contains the value of a dict. CPython and pypy serialise the fields in different orders, so we can't assert on the whole string.

rudigiesler:
:+1:

PR #892, opened 2015-01-07
Add a SetOf model field type

jerith:
We need a field type that gives us set-like behaviour, but is stored as a JSON list under the hood.


justinvdm:
:+1: from me once Travis is happy.

PR #893, opened 2015-01-12
Tools for extracting message store stats

jerith:
We specifically want to get at the unique address counts, but breakdowns of various other stats by time period would also be useful and can probably be implemented together more efficiently than separately.

### DIFF
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -255,10 +255,7 @@ class MessageStore(object):
                 old_key = key_manager.add_key(key, timestamp)
                 if old_key is not None:
                     key_count += 1
-            if index_page.has_next_page():
-                index_page = yield index_page.next_page()
-            else:
-                index_page = None
+            index_page = yield index_page.next_page()
 
         yield self.cache.add_inbound_message_count(batch_id, key_count)
         for key, timestamp in key_manager:
@@ -287,10 +284,7 @@ class MessageStore(object):
                     sc = yield self.get_event_counts(old_key[0])
                     for status, count in sc.iteritems():
                         status_counts[status] += count
-            if index_page.has_next_page():
-                index_page = yield index_page.next_page()
-            else:
-                index_page = None
+            index_page = yield index_page.next_page()
 
         yield self.cache.add_outbound_message_count(batch_id, key_count)
         for status, count in status_counts.iteritems():
@@ -318,10 +312,7 @@ class MessageStore(object):
                 status_counts[status] += 1
                 if status.startswith("delivery_report."):
                     status_counts["delivery_report"] += 1
-            if index_page.has_next_page():
-                index_page = yield index_page.next_page()
-            else:
-                index_page = None
+            index_page = yield index_page.next_page()
 
         returnValue(status_counts)
 
@@ -943,6 +934,8 @@ class KeysWithTimestamps(object):
             of results.
         """
         next_page = yield self._index_page.next_page()
+        if next_page is None:
+            returnValue(None)
         returnValue(type(self)(self._message_store, self._batch_id, next_page))
 
     def has_next_page(self):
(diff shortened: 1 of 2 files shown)


smn:
:+1: from me but what's the reasoning behind total & unique? Does total reflect total number of messages received?

Is there value in being able to retrieve the uniques rather than just the `len()` ?

jerith:
> what's the reasoning behind total & unique? Does total reflect total number of messages received?

Yes. We need to walk the whole index anyway to get the uniques, so we may as well count the messages while we're at it. It costs us nothing and will avoid looking up the total separately for things that needs both.

jerith:
> Is there value in being able to retrieve the uniques rather than just the `len()` ?

Not that I can think of. If you need to do something with that, you're probably better off exporting the messages and processing them yourself.

vumi/components/message_store.py
```diff
@@ -813,6 +813,108 @@ def message_event_keys_with_statuses(self, msg_id, max_results=None):
...
+        page = KeysWithAddresses(self, batch_id, raw_page)
+
+        while page is not None:
+            results = list(page)
+            total += len(results)
+            unique_addresses.update(addr for key, timestamp, addr in results)
+            if page.has_next_page():
+                page = yield page.next_page()
+            else:
+                page = None
```
hodgestar: Should we update this not that #901 has landed?

vumi/components/message_store.py
```diff
@@ -813,6 +813,108 @@ def message_event_keys_with_statuses(self, msg_id, max_results=None):
...
+            max_results = self.DEFAULT_MAX_RESULTS
+        raw_page = yield self.inbound_messages.index_keys_page(
+            'batches_with_addresses', start_value, end_value,
+            return_terms=True, max_results=max_results)
+        page = KeysWithAddresses(self, batch_id, raw_page)
+
+        while page is not None:
+            results = list(page)
+            total += len(results)
+            unique_addresses.update(addr for key, timestamp, addr in results)
```
hodgestar: A little worried this builds up a fairly big set in RAM for batches with lots of unique addresses?

vumi/components/message_store.py
```diff
@@ -813,6 +813,108 @@ def message_event_keys_with_statuses(self, msg_id, max_results=None):
...
+            if page.has_next_page():
+                page = yield page.next_page()
+            else:
+                page = None
```
hodgestar: And here.

hodgestar:
Left some small comments. Looks good to me.

Regarding building up big sets in RAM: If this becomes a problem in practice, perhaps we can swap out the ``set`` for something like https://github.com/svpcom/hyperloglog (or even our own implementation if we don't like any of the existing ones) so I don't think we necessarily let that comment hold up this PR.

For publishing this metrics in Vumi Go, I'm thinking of something like a simple Celery task that goes through active conversations once a day and calculates metrics for that day.

vumi/components/message_store.py
jerith: We should, thanks.

jerith:
> For publishing this metrics in Vumi Go, I'm thinking of something like a simple Celery task that goes through active conversations once a day and calculates metrics for that day.

I'd be much happier doing this on-demand, both because it's more flexible (in that one can select specific time ranges) and because we avoid doing a bunch of unnecessary work.

A daily job that walks indexes for every active conversation doesn't scale very with either the number of conversations or the number of messages in the conversations. (We can look at finding a way of doing this if the majority of account owners really need low-latency access to this information on a daily basis, but I'd be surprised if that were the case.)

jerith:
> Regarding building up big sets in RAM: If this becomes a problem in practice, perhaps we can swap out the `set` for something like https://github.com/svpcom/hyperloglog (or even our own implementation if we don't like any of the existing ones) so I don't think we necessarily let that comment hold up this PR.

Agreed, but there isn't an easy way to do this both accurately and cheaply as far as I'm aware. (That's another reason I'd prefer to do it on-demand rather than daily.)

hodgestar:
:+1:

PR #894, opened 2015-01-14
SMPP transport fails to send empty message

jerith:
Specifically, it can't encode `None`. We can fall back to an empty string instead.

There's probably no reason for us to ever send empty messages, but we can receive them without any problems and there doesn't seem to be any reason to forbid them either.


rudigiesler:
:+1:

PR #895, opened 2015-01-20
add basic address type

rudigiesler:


vumi/message.py
```diff
@@ -219,6 +231,8 @@ def process_fields(self, fields):
         fields.setdefault('content', None)
         fields.setdefault('transport_metadata', {})
         fields.setdefault('group', None)
+        fields.setdefault('to_addr_type', self.AT_MSISDN)
+        fields.setdefault('from_addr_type', self.AT_MSISDN)
```
hodgestar: I wonder if we should allow ``None`` to be a value and have it mean ``unspecified``?

vumi/message.py
```diff
@@ -210,6 +210,18 @@ class TransportUserMessage(TransportMessage):
     TRANSPORT_TYPES = set([TT_HTTP_API, TT_IRC, TT_TELNET, TT_TWITTER, TT_SMS,
                            TT_USSD, TT_XMPP, TT_MXIT, TT_WECHAT])
 
+    AT_IRC_NICKNAME = 'irc_nickname'
+    AT_TELNET_ADDRESS = 'telnet_address'
```
hodgestar: I'm not sure whether it makes much sense to have ``AT_TELNET_ADDRESS``.

vumi/message.py
```diff
@@ -243,6 +257,8 @@ def validate_fields(self):
         if self['session_event'] not in self.SESSION_EVENTS:
             raise InvalidMessageField("Invalid session_event %r"
                                       % (self['session_event'],))
+        self.assert_field_value('from_addr_type', *self.ADDRESS_TYPES)
+        self.assert_field_value('to_addr_type', *self.ADDRESS_TYPES)
```
hodgestar: I'm wondering whether we want to force these to be in a specific set? It seems more like something a transport should check (i.e. is this address type one that I support).

vumi/message.py
rudigiesler: I wasn't sure what the to and from addresses represent for telnet. IP address? Port?

vumi/message.py
jerith: Agreed. We can let the thing handling the message decide what a sensible default is.

vumi/message.py
justinvdm: Seconded

vumi/message.py
jerith: IP and port. We can probably leave it out -- the telnet transport is only really useful for local testing and it can pretend to be some other address type if necessary.

hodgestar:
:+1:

We should check that Vumi Go tests pass with this.

rudigiesler:
hodgestar: All the Vumi Go tests pass locally for me. Is there something more we can do to test?

hodgestar:
Nope, I think that's good.

PR #896, opened 2015-01-23
add priority for middleware

rudigiesler:
Add the ability to specify a `consume_priority` and `publish_priority` to each middleware when describing their order in the config files.

Middlewares are executed in order for consume, and reverse order for publish. All middlewares that have priority will be executed first, in priority order. Order will be in order for consume and reverse order for publish for middleware with the same priority level

### DIFF
--- a/vumi/middleware/base.py
+++ b/vumi/middleware/base.py
@@ -239,14 +239,16 @@ def create_middlewares_from_config(worker, config):
         middleware_config = config.get(middleware_name, {})
         if type(item[middleware_name]) is str:
             cls_name = item[middleware_name]
+            middleware_config['consume_priority'] = 0
+            middleware_config['publish_priority'] = 0
         elif type(item[middleware_name]) is dict:
             conf = item[middleware_name]
             cls_name = conf.get('class')
             try:
                 middleware_config['consume_priority'] = int(conf.get(
-                    'consume_priority', None))
+                    'consume_priority', 0))
                 middleware_config['publish_priority'] = int(conf.get(
-                    'publish_priority', None))
+                    'publish_priority', 0))
             except ValueError:
                 raise ConfigError(
                     "Middleware priority level must be an integer")
--- a/vumi/middleware/base.py
+++ b/vumi/middleware/base.py
@@ -184,12 +184,10 @@ class MiddlewareStack(object):
 
     @staticmethod
     def _sort_by_priority(middlewares, priority_key):
-        priorities = defaultdict(list)
-        for mw in middlewares:
-            priorities[mw.config.get(priority_key, None)].append(mw)
-        sorted_priorities = sorted(
-            priorities.keys(), key=lambda a: '' if not a else a)
-        return [mw for p in sorted_priorities for mw in priorities[p]]
+        sorted_middlewares = sorted(
+            (mw.config[priority_key], idx, mw)
+            for idx, mw in enumerate(middlewares))
+        return [mw for priority, idx, mw in sorted_middlewares]
 
     @inlineCallbacks
     def _handle(self, middlewares, handler_name, message, connector_name):
--- a/vumi/middleware/base.py
+++ b/vumi/middleware/base.py
@@ -187,8 +187,8 @@ class MiddlewareStack(object):
         priorities = defaultdict(list)
         for mw in middlewares:
             priorities[mw.config.get(priority_key, None)].append(mw)
-        sorted_priorities = [p for p in sorted(
-            priorities.keys(), key=lambda a: '' if not a else a)]
+        sorted_priorities = sorted(
+            priorities.keys(), key=lambda a: '' if not a else a)
         return [mw for p in sorted_priorities for mw in priorities[p]]
 
     @inlineCallbacks
--- a/vumi/middleware/base.py
+++ b/vumi/middleware/base.py
@@ -1,4 +1,5 @@
 # -*- test-case-name: vumi.middleware.tests.test_base -*-
+from collections import defaultdict
 
 from twisted.internet.defer import inlineCallbacks, returnValue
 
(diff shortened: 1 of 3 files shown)


rudigiesler:
Currently, specifying middleware looks like this:

```
middleware:
  - logging_mw: vumi.middleware.logging.LoggingMiddleware
  - storing_mw: vumi.middleware.message_storing.StoringMiddleware
```

with this PR, the old way is still valid, but now you can specify priority like this:

```
middleware:
  - logging_mw: {
        class: vumi.middleware.logging.LoggingMiddleware,
        consume_priority: 1,
        publish_priority: 3
    }
```

vumi/middleware/tests/test_base.py
```diff
@@ -89,6 +89,16 @@ def mkmiddleware(self, name, mw_class):
         yield mw.setup_middleware()
         returnValue(mw)
 
+    @inlineCallbacks
+    def mk_priority_middleware(self, name, mw_class, consume_pri, publish_pri):
+        mw = mw_class(
+            name, {
+                'consume_priority': consume_pri,
+                'publish_priority': publish_pri
```
jerith: Trailing comma, please.

vumi/middleware/tests/test_base.py
```diff
@@ -225,3 +267,16 @@ def test_parse_yaml(self):
...
+        priority2 = ToyMiddleware('priority2', {"priority": 2}, self)
+        nopriority = ToyMiddleware('nopriority', {}, self)
+        priority1_1 = ToyMiddleware('priority1_1', {"priority": 1}, self)
+        priority1_2 = ToyMiddleware('priority1_2', {"priority": 1}, self)
+        middlewares = [priority2, nopriority, priority1_1, priority1_2]
+        for mw in middlewares:
+            yield mw.setup_middleware()
+        mw_sorted = MiddlewareStack._sort_by_priority(middlewares, 'priority')
+        self.assertEqual(
+            mw_sorted, [priority1_1, priority1_2, priority2, nopriority])
```
jerith: Missing newline.

jerith:
If priorities are expected to be integers, we should validate that they are. Also, I'd prefer an explicit default priority -- at the moment it's unclear how prioritised and unprioritised middlewares interact and I'm not sure the set of possible priorities includes both sides of the unprioritised group.

vumi/middleware/base.py
```diff
@@ -176,7 +177,19 @@ class MiddlewareStack(object):
     """
 
     def __init__(self, middlewares):
-        self.middlewares = middlewares
+        self.consume_middlewares = self._sort_by_priority(
+            middlewares, 'consume_priority')
+        self.publish_middlewares = self._sort_by_priority(
+            reversed(middlewares), 'publish_priority')
```
justinvdm: I'm wondering whether we should make the two priorities work in ascending order instead of ascending for the one and ascending for the other, seems a bit surprising. It would mean breaking backwards compatibility I guess, though, but maybe that's okay,

jerith:
Setting an explicit default priority would also let us simplify the sorting code by building a list of (priority, index_from_config, middleware) which we can sort directly instead of the two-step process we have now.

justinvdm:
+1 on a default priority and using it as the first sorting key and the index as the secondary sorting key. The middleware config is a dict instead of a list though, so not sure if getting consitent indices would be possible.

justinvdm:
Oh, woops, it is actually a list, so that would be possible.

vumi/middleware/base.py
```diff
@@ -176,7 +177,19 @@ class MiddlewareStack(object):
...
+
+    @staticmethod
+    def _sort_by_priority(middlewares, priority_key):
+        priorities = defaultdict(list)
+        for mw in middlewares:
+            priorities[mw.config.get(priority_key, None)].append(mw)
+        sorted_priorities = sorted(
+            priorities.keys(), key=lambda a: '' if not a else a)
+        return [mw for p in sorted_priorities for mw in priorities[p]]
```
hodgestar: I wonder if we could just replace this function with ``sorted(middlewares, key=lambda mw: mw.config.get(priority_key, ''))`` and rely on Python's sort being a stable sort?

hodgestar:
Is there a good use case for having middlewares that are only on either publish or consume or only on particular endpoints or connectors?

vumi/middleware/base.py
jerith: We should still have the same behaviour if the configs don't specify priorities, otherwise we need to update all middleware configs in lockstep with the code change.

vumi/middleware/base.py
```diff
@@ -214,14 +225,37 @@ def create_middlewares_from_config(worker, config):
...
+                " key should be a name for the middleware. The value should be"
+                " the full dotted name of the class implementing the"
+                " middleware, or a mapping containing the keys 'class' with a"
+                " value of the full dotted class name, 'consume_priority' with"
+                " the priority level for consuming, and 'publish_priority'"
+                " with the priority level for publishing, both integers.")
         middleware_name = keys[0]
-        cls_name = item[middleware_name]
         middleware_config = config.get(middleware_name, {})
+        if type(item[middleware_name]) is str:
```
jerith: Rather use `isinstance()` (and `basestring` instead of `str`) than directly comparing types.

hodgestar:
My one slight criticism of the current approach is that it's now hard to tell what the ordering is by looking at just the ``middleware`` configuration section. One has to also read all of the individual middleware config sections.

Anyone have any good ideas on how we might structure the config to avoid this? If not, don't let this hold things up.

Looks good otherwise.

justinvdm:
@hodgestar One other approach that might work would be to do something like:

```yaml
mw_publish_order:
  - mw1
  - mw2
  - mw3

mw_consume_order:
  - mw1
  - mw3
  - mw2
```

The advantage is that the ordering is more clear, the disadvantage is that priorities are set much less explicitly.

Happy with the current approach though.

hodgestar:
I was completely wrong about how the config functioned. Doh. My issue is already addressed.

:+1: from me.

justinvdm:
:+1:

vumi/middleware/base.py
```diff
@@ -176,7 +175,14 @@ class MiddlewareStack(object):
...
     def __init__(self, middlewares):
-        self.middlewares = middlewares
+        self.consume_middlewares = self._sort_by_priority(
+            middlewares, 'consume_priority')
+        self.publish_middlewares = self._sort_by_priority(
+            reversed(middlewares), 'publish_priority')
+
+    @staticmethod
+    def _sort_by_priority(middlewares, priority_key):
+        return sorted(middlewares, key=lambda mw: mw.config[priority_key])
```
jerith: Can you add a comment here about relying on Python's sorting algorithm being stable? It's not something that's likely to change, but it will help us understand how the ordering works when we look at this code again later.

jerith:
One comment about explaining the sorting, otherwise :+1:

jerith:
:+1:

PR #899, opened 2015-01-27
FakeRedis ZREMRANGEBYRANK behaves differently from Redis

jerith:
Specifically, it seems to remove the wrong number of entries in some cases.

### DIFF
--- a/vumi/persist/tests/test_fake_redis.py
+++ b/vumi/persist/tests/test_fake_redis.py
@@ -164,6 +164,14 @@ class TestFakeRedis(VumiTestCase):
             [('three', 3)], 'zrange', 'set', 0, -1, withscores=True)
 
     @inlineCallbacks
+    def test_zremrangebyrank_empty_range(self):
+        yield self.redis.zadd('set', one=1, two=2, three=3)
+        yield self.assert_redis_op(0, 'zremrangebyrank', 'set', 10, 11)
+        yield self.assert_redis_op(
+            [('one', 1), ('two', 2), ('three', 3)],
+            'zrange', 'set', 0, -1, withscores=True)
+
+    @inlineCallbacks
     def test_zremrangebyrank_negative_start(self):
         yield self.redis.zadd('set', one=1, two=2, three=3)
         yield self.assert_redis_op(2, 'zremrangebyrank', 'set', -2, 2)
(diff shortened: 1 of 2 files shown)


vumi/persist/tests/test_fake_redis.py
```diff
@@ -178,6 +186,14 @@ def test_zremrangebyrank_negative_stop(self):
             [('one', 1)], 'zrange', 'set', 0, -1, withscores=True)
 
     @inlineCallbacks
+    def test_zremrangebyrank_negative_stop_empty_range(self):
+        yield self.redis.zadd('set', one=1, two=2, three=3)
+        yield self.assert_redis_op(0, 'zremrangebyrank', 'set', 0, -5)
```
justinvdm: Would it be worth having a test for a start length greater than the length of the set also? If I'm understanding things correctly, that case would also have failed before the fix in this PR.

jerith: That case wasn't failing (it was just the negative numbers we were mishandling), but it's worth having a test for it anyway. Thanks.

justinvdm: Sorry, I should've been more specific, I meant negative numbers whose absolute value is greater than the length of the set, like we are doing here for the stop param with `-5`.

jerith: Ah, right. The problem here was that we were mishandling all negative values, so a single test case each for the combinations of empty and non-empty ranges with a negative value at each end is sufficient.

justinvdm: I was thinking of a case where we do something like:

```
ZREMRANGEBYRANK set -5 -4
```

(where `set` has a length of `3`).

With [the current way](https://github.com/praekelt/vumi/blob/develop/vumi/persist/fake_redis.py#L506-L507) we work with negative start values, I think we would normalize `-5` to `-2` and `-4` to `-1`, so we would end up removing a non-empty range of values. I thought it might be useful to add a test that fails for the current implementation and passes for the fix in this PR.

I think the current tests are sufficient for checking how we handle both empty and non-empty ranges when we use negative start and end values, but just thought it might be worth adding a test for something that is currently failing. Not very important though, your call really.

justinvdm:
:+1: when travis is happy

PR #898, opened 2015-01-27
add yaml import constructor

rudigiesler:
There is no way to import yaml files from within yaml files. This would be useful for things like not having to repeat a list of provider prefixes for every transport, and instead listing a file in each of the transports, and having a single copy of that file.

I propose to do this by lightly wrapping the pyyaml `SafeLoader` class, and adding an `import` directive that will `safe_load` the specified file.

### DIFF
--- a/vumi/servicemaker.py
+++ b/vumi/servicemaker.py
@@ -1,5 +1,5 @@
 # -*- test-case-name: vumi.tests.test_servicemaker -*-
-
+import os
 import sys
 import warnings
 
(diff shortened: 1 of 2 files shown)


hodgestar:
For the record, the difficulty I encountered previously was the inability to reference sections from an included file (e.g. ``&COMMON_SETTINGS``) but that doesn't affect this use case.

justinvdm:
:+1:

hodgestar:
:+1:

PR #901, opened 2015-01-30
Better VumiIndexPage.next_page()

jerith:
To avoid having to check `.has_next_page()` all the time, `.next_page()` should return `None` on the last page.


hodgestar:
:+1: but Travis builds are currently failing -- it looks like maybe something changed in Twisted?

jerith:
See #902 for the test fixes.

hodgestar:
:+1: once Travis is happy again.

PR #904, opened 2015-02-05
Message store listing script

jerith:


vumi/scripts/vumi_list_messages.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class Options(usage.Options):
+    optParameters = [
+        ["batch", None, None,
+         "Batch identifier to list messages for."],
+        ["bucket-prefix", "b", None,
+         "The bucket prefix for the Riak manager."],
+        ["direction", None, None,
+         "Message direction. Valid values are `inbound' and `outbound'."],
```
hodgestar: I think there are single-quotes here instead of backticks. Or perhaps the other way around?

vumi/scripts/vumi_list_messages.py
```diff
@@ -0,0 +1,97 @@
...
+        ["index-page-size", None, "1000",
+         "The number of keys to fetch in each index query."],
+    ]
+
+    longdesc = """
+    Index-based message store lister. For each message, the timestamp, remote
+    address, and message_id are returned in a comma-separated format.
+    """
+
+    def ensure_dependent_option(self, needed, needs):
```
hodgestar: Function appears to be unused.

vumi/scripts/vumi_list_messages.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def get_riak_manager(self, riak_config):
+        return TxRiakManager.from_config(riak_config)
+
+    def emit(self, s):
+        print s
+
+    @inlineCallbacks
+    def list_pages(self, index_page):
+        while index_page is not None:
```
hodgestar: Should we output a header row?

hodgestar:
Left a few comments. Looks good otherwise.

vumi/scripts/vumi_list_messages.py
jerith: Thanks. I copied it from elsewhere and then forgot to delete it when I deleted the things that used it.

vumi/scripts/vumi_list_messages.py
jerith: I'd rather not, because we may want to pipe the output of this into something else. I'm not opposed to adding an option to output one, though.

vumi/scripts/vumi_list_messages.py
jerith: There are a bunch of things that use starting-backtick-ending-apostrophe for quoting values like that, especially in usage and manpages. I can change it to apostrophes on both sides if you prefer.

hodgestar:
:+1:

PR #905, opened 2015-02-06
Riak 1.4 in Travis

jerith:
Travis switched to Riak 2.0 in their build images, which breaks our builds.


smn:
The only thing I'd add is:

```yaml
before_script:
  - riak version
  - riak-admin member-status
```

to have that info in the log as part of the test run so you can see what version you're running.

smn:
:+1: from me.

PR #906, opened 2015-02-09
Add paginated index queries to foreign key backlinks

jerith:
We currently do unpaginated index queries, which is a problem.

### DIFF
--- a/vumi/persist/fields.py
+++ b/vumi/persist/fields.py
@@ -729,18 +729,35 @@ class ForeignKeyDescriptor(FieldDescriptor):
             self.index_name = "%s_bin" % self.key
         else:
             self.index_name = self.field.index
-        if self.field.backlink is None:
-            self.backlink_name = model_cls.__name__.lower() + "s"
-        else:
-            self.backlink_name = self.field.backlink
+
+        backlink_name = self.field.backlink
+        if backlink_name is None:
+            backlink_name = model_cls.__name__.lower() + "s"
         self.other_model.backlinks.declare_backlink(
-            self.backlink_name, self.reverse_lookup_keys)
+            backlink_name, self.reverse_lookup_keys)
+
+        backlink_keys_name = backlink_name + "_keys"
+        if backlink_keys_name.endswith("s_keys"):
+            backlink_keys_name = backlink_name[:-1] + "_keys"
+        self.other_model.backlinks.declare_backlink(
+            backlink_keys_name, self.reverse_lookup_keys_paginated)
 
     def reverse_lookup_keys(self, modelobj, manager=None):
         if manager is None:
             manager = modelobj.manager
         return manager.index_keys(
             self.model_cls, self.index_name, modelobj.key)
+
+    def reverse_lookup_keys_paginated(self, modelobj, manager=None,
+                                      max_results=None, continuation=None):
+        """
+        Perform a paginated index query for backlinked objects.
+        """
+        if manager is None:
+            manager = modelobj.manager
+        return manager.index_keys_page(
+            self.model_cls, self.index_name, modelobj.key,
+            max_results=max_results, continuation=continuation)
 
     def clean(self, modelobj):
         if self.key not in modelobj._riak_object.get_data():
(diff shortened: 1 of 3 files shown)


vumi/persist/fields.py
```diff
@@ -729,12 +729,18 @@ def setup(self, model_cls):
...
+        backlink_name = self.field.backlink
+        if backlink_name is None:
+            backlink_name = model_cls.__name__.lower() + "s"
         self.other_model.backlinks.declare_backlink(
-            self.backlink_name, self.reverse_lookup_keys)
+            backlink_name, self.reverse_lookup_keys)
+
+        backlink_keys_name = backlink_name + "_keys"
+        if backlink_keys_name.endswith("s_keys"):
+            backlink_keys_name = backlink_name[:-1] + "_keys"
```
rudigiesler: wouldn't `backlink_keys_name = backlink_name.rstrip('s') + '_keys'` be more concise than these 3 lines?

rudigiesler: Nevermind, that would strip all trailing `s` characters.

justinvdm: Will this work for cases where `self.field.backlink` ends with an 's'?

justinvdm: For example, if `self.field.backlink` was `'boss'`, `backlink_keys_name` would in up being `'bos_keys'`. Is that fine?

jerith: Not really, but fixing that is a deep rabbit hole that I didn't want to dive into here. It's not that big a deal, though, because we're not using `field.backlink` anywhere that I can find.

vumi/persist/tests/test_model.py
```diff
@@ -1197,6 +1202,11 @@ def test_reverse_foreignkey_fields(self):
         results = yield s2.backlinks.foreignkeymodels()
         self.assertEqual(sorted(results), ["bar1", "bar2"])
 
+        s2 = yield simple_model.load("foo")
+        results = yield s2.backlinks.foreignkeymodel_keys()
+        self.assertEqual(sorted(results), ["bar1", "bar2"])
+        self.assertEqual(results.has_next_page(), False)
```
justinvdm: Should we maybe test that pagination is working properly here: using `max_results` to limit the returned keys, checking the returned keys, checking that `has_next_page()` is `True`, using `.next_page()`, checking the returned keys and checking that `has_next_page()` is `False`?

Or is there a reason why just a single `has_next_page()` check is sufficient?

justinvdm:
Bikeshed: Without reading code or docstrings, it isn't clear that `.backlinks.foo()` returns the non-paginated keys and `.backlinks.foo_keys()` returns paginated keys. Shouldn't we maybe make that explicit (`backlinks.foo_paged()` or something)?

vumi/persist/tests/test_model.py
jerith: The `has_next_page()` check is there to make sure we get something that looks like an index page. The actual index page implementations are tested elsewhere and don't need to be tested again here.

jerith:
> Bikeshed: Without reading code or docstrings, it isn't clear that `.backlinks.foo()` returns the non-paginated keys and `.backlinks.foo_keys()` returns paginated keys. Shouldn't we maybe make that explicit (`backlinks.foo_paged()` or something)?

I want to eventually get rid of all non-paginated index results, at which point `foo_paged()` won't make much sense. `foo_keys()` already makes more sense than `foos()`, which implies that you get model objects rather than keys.

vumi/persist/fields.py
justinvdm: I would've thought it is as simple enough to avoid:

```python
if self.field.backlink is None:
    name = model_cls.__name__.lower()
    backlink_name = name + "s"
    backlink_keys_name = name + "_keys"
else:
    backlink_name = self.field.backlink
    backlink_keys_name = self.field.backlink + "_keys"

self.other_model.backlinks.declare_backlink(
    backlink_name, self.reverse_lookup_keys)

self.other_model.backlinks.declare_backlink(
    backlink_keys_name, self.reverse_lookup_keys_paginated)
```

What is the rabbit hole? Are we doing this somewhere else too?

justinvdm:
> I want to eventually get rid of all non-paginated index results, at which point foo_paged() won't make much sense. foo_keys() already makes more sense than foos(), which implies that you get model objects rather than keys.

+1, happy with that.

vumi/persist/fields.py
jerith: That means that if you specify `field.backlink`, you either get `foo` and `foo_keys` or `foos` and `foos_keys`.

justinvdm: That seems less surprising than finding out getting either `foo` and `foo_keys` or `foos` and `foo_keys` (I would think 'what happened to my 's'?). I guess this is bikesheddy enough to leave as is, but it still does seem like surprising behaviour.

justinvdm:
:+1:

justinvdm:
:+1: again, when travis is happy.

PR #907, opened 2015-02-12
add session length middleware

rudigiesler:


vumi/middleware/session_length.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @inlineCallbacks
+    def teardown_middleware(self):
+        yield self.redis.close_manager()
+
+    def handle_inbound(self, message, connector_name):
+        redis_key = '%s:%s' % (message.get('from_addr'), 'session_created')
+        if message.get('event_type') == SESSION_NEW:
+            yield self.redis.set(redis_key, str(time.time()))
+        elif message.get('event_type') == SESSION_CLOSE:
```
hodgestar: Should we also have a ``handle_outbound`` that checks for ``SESSION_NEW`` and ``SESSION_CLOSE`` (close is more important, but in theory new is also possible on outbound).

rudigiesler: Yes, this PR is not quite finished yet. I'm busy with tests first, then I'll do the same for outbound (and maybe split this out into a function that outbound and inbound can call)

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,75 @@
...
+        r_config = self.config.get('redis_manager', {})
+        self.redis = yield TxRedisManager.from_config(r_config)
+        self.timeout = self.config.get('timeout', 120)
+
+    @inlineCallbacks
+    def teardown_middleware(self):
+        yield self.redis.close_manager()
+
+    def _set_message_billing_metadata(self, message, field, value):
+        if not message['helper_metadata'].get('billing'):
```
justinvdm: Do we maybe want to make the field name used in `helper_metadata` configurable?

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def _set_session_start_time(self, message, time):
+        self._set_message_billing_metadata(message, 'session_start', time)
+
+    def _set_session_end_time(self, message, time):
+        self._set_message_billing_metadata(message, 'session_end', time)
+
+    @inlineCallbacks
+    def _process_message(self, message, redis_key):
+        if message.get('event_type') == self.SESSION_NEW:
```
justinvdm: I think we want to use the `session_event` field. I think `even_type` is a field used for event messages (acks, nacks and delivery reports).

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,75 @@
...
+            start_time = time.time()
+            yield self.redis.setex(redis_key,  self.timeout, str(start_time))
+            self._set_session_start_time(message, start_time)
+        elif message.get('event_type') == self.SESSION_CLOSE:
+            self._set_session_end_time(message, time.time())
```
justinvdm: We use `str()` for the start time, do we maybe want to use the same here?

justinvdm: Oh, nevermind, this one doesn't go in redis.

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,83 @@
...
+                self._set_session_start_time(message, float(created_time))
+                yield self.redis.delete(redis_key)
+        else:
+            created_time = yield self.redis.get(redis_key)
+            if created_time:
+                self._set_session_start_time(message, float(created_time))
+
+    @inlineCallbacks
+    def handle_inbound(self, message, connector_name):
+        redis_key = '%s:%s' % (message.get('from_addr'), 'session_created')
```
justinvdm: I think we might want to namespace the keys a bit more to avoid collisions between transports with a message from/to the same address. In the [metrics middleware](https://github.com/praekelt/vumi-go/blob/develop/go/vumitools/middleware.py#L310-L313), we seem to make use of the transport name, maybe we should do that here?

Separate thing -- I don't want to bikeshed this too much, but it might be worth refactoring how we generate the key into a method so we don't repeat ourselves.

vumi/middleware/tests/test_session_length.py
```diff
@@ -0,0 +1,141 @@
...
+        self.assertTrue(isinstance(
+                msg['helper_metadata']['session']['session_start'], float))
+
+        msg = yield mw.handle_inbound(msg_end, "dummy_connector")
+        value = yield self.redis.get('+54321:session_created')
+        self.assertTrue(value is None)
+        self.assertTrue(isinstance(
+                msg['helper_metadata']['session']['session_start'], float))
+        self.assertTrue(isinstance(
+            msg['helper_metadata']['session']['session_end'], float))
```
justinvdm: I think it might be worth stubbing `time.time` with `self.patch` to check that we are storing the right values for the start and end.

justinvdm:
Left some minor comments, looks good :)

justinvdm:
:+1: when travis is happy

justinvdm:
(but lets land on Monday)

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    SESSION_NEW, SESSION_CLOSE = (
+        TransportUserMessage.SESSION_NEW, TransportUserMessage.SESSION_CLOSE)
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        r_config = self.config.get('redis_manager', {})
+        self.redis = yield TxRedisManager.from_config(r_config)
+        self.timeout = self.config.get('timeout', 120)
+        self.field_name = self.config.get('field_name', 'session')
```
hodgestar: We should perhaps we using a proper config class for these (happy if that happens in a separate PR).

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,87 @@
...
+        self.clock = reactor
+
+    @inlineCallbacks
+    def teardown_middleware(self):
+        yield self.redis.close_manager()
+
+    def _set_message_billing_metadata(self, message, field, value):
```
hodgestar: Nothing else mentions ``billing`` so perhaps this method should be ``_set_message_session_metadata``?

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,87 @@
...
+    in the session, as well as the end timestamp if the session in
+    message['helper_metadata'][field_name]['session_end'] if the message
+    marks the end of the session.
+
+    Configuration option:
+
+    :param dict redis:
+        Redis configuration parameters.
+    :param int timeout:
+        Redis key timeout (in seconds). Defaults to 120.
```
hodgestar: Session can be a bit longer than this so I'd prefer a timeout of, say, ``600`` seconds?

hodgestar:
Left some really minor comments. Happy for them to be addressed in a separate PR or skipped.

vumi/middleware/session_length.py
rudigiesler: I didn't see a config class used in any of the other vumi middlewares, they were all configured like this, which is why I did the same here. Happy to switch over in a separate PR.

rudigiesler: PR create here: https://github.com/praekelt/vumi/pull/910

justinvdm:
:+1:

PR #908, opened 2015-02-12
Change provider setting middleware to only set the provider if it doesn't exist

justinvdm:
This will ensure we don't overwrite a provider set by something else (like a transport).


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #909, opened 2015-02-12
SMPP transport throttling should handle reconnects

jerith:


vumi/transports/smpp/smpp_transport.py
```diff
@@ -499,6 +504,12 @@ def _check_stop_throttling(self):
         """
         self._unthrottle_delayedCall = None
 
+        if not self.service.is_bound():
+            # We don't have a bound SMPP connection, so try again later.
+            log.msg("Can't check throttling while unbound, trying later.")
+            self.check_stop_throttling(self.get_static_config().throttle_delay)
```
justinvdm: Does this mean we check to see if we should stop throttling again after `throttle_delay` seconds if the service isn't bound?

justinvdm: If that is the case, maybe we should mention something about how it is used when we are not yet bound in [the config field's description](https://github.com/praekelt/vumi/blob/feature/issue-909-smpp-throttling-reconnect/vumi/transports/smpp/config.py#L20-L22)?

justinvdm:
Left a minor comment, but looks good. Not sure if someone with more SMPP battle scars should take a look though.

hodgestar:
:+1:

vumi/transports/smpp/smpp_transport.py
jerith: This is only relevant during a reconnect (we can't be throttled until we've sent at least one message), and it has more or less the same behaviour as retrying a throttled message and getting another RTHROTTLED response.

PR #910, opened 2015-02-16
Create config class for session length middleware

rudigiesler:


rudigiesler:
I'm wondering if we should use this opportunity to update all of the middlewares to use a config class?

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,101 @@
...
+
+class SessionLengthMiddlewareConfig(Config):
+    """
+    Configuration class for the session length middleware.
+    """
+
+    redis = ConfigDict("Redis config", default={})
+    timeout = ConfigInt("Redis key timeout (secs)", default=600)
+    field_name = ConfigText(
+        "Field name in message helper_metadata", default="session")
```
justinvdm: Could we make these fields static?

vumi/middleware/session_length.py
```diff
@@ -0,0 +1,101 @@
...
+    :param str field_name:
+        The field name to use when storing the timestamps in the message
+        helper_metadata. Defaults to 'session'.
+    """
+    SESSION_NEW, SESSION_CLOSE = (
+        TransportUserMessage.SESSION_NEW, TransportUserMessage.SESSION_CLOSE)
+
+    @inlineCallbacks
+    def setup_middleware(self):
+        config = SessionLengthMiddlewareConfig(self.config)
```
justinvdm: Could we use `static=True` in the constructor to only get the static config fields?

justinvdm:
Left minor comments, looks good.

vumi/middleware/manhole.py
```diff
@@ -11,6 +14,14 @@
 from twisted.internet.endpoints import serverFromString
 
 
+class ManholeMiddlewareConfig(Config):
+    twisted_enpoint = ConfigText(
```
jerith: Missing `d`.

jerith: Also, I think we have a specific field type for these.

rudigiesler: Would it be the URL field type? https://github.com/praekelt/confmodel/blob/develop/confmodel/fields.py#L71

hodgestar: I think it's ``ConfigServerEndpoint`` in ``vumi.config``.

vumi/middleware/logging.py
```diff
@@ -1,9 +1,24 @@
 # -*- test-case-name: vumi.middleware.tests.test_logging -*-
 
+from confmodel import Config
+from confmodel.fields import ConfigText
+
 from vumi.middleware import BaseMiddleware
 from vumi import log
 
 
+class LoggingMiddlewareConfig(Config):
```
hodgestar: These should be extending ``BaseMiddlewareConfig``, I think?

vumi/middleware/base.py
```diff
@@ -29,12 +37,16 @@ class BaseMiddleware(object):
 
     If you are subclassing this class, you should not override
     :meth:`__init__`. Custom setup should be done in
-    :meth:`setup_middleware` instead.
+    :meth:`setup_middleware` instead. The config class can be overidden by
+    replacing the ``config_class`` class variable.
     """
+    config_class = BaseMiddlewareConfig
```
hodgestar: Perhaps this should be ``CONFIG_CLASS`` for consistency with applications and transports?

vumi/middleware/base.py
```diff
@@ -29,12 +37,16 @@ class BaseMiddleware(object):
...
 
     def __init__(self, name, config, worker):
         self.name = name
-        self.config = config
+        self.config = self.config_class(config)
```
hodgestar: I think this should follow the pattern applications and transports use, i.e. ``self.config = self.CONFIG_CLASS(self.config, static=True)``. This would mean all middleware config is static, which I think is fine (having it depend on the message would just be complicated).

vumi/middleware/message_storing.py
```diff
@@ -9,6 +12,30 @@
 from vumi.persist.txredis_manager import TxRedisManager
 
 
+class StoringMiddlewareConfig(Config):
+    """
+    Config class for the storing middleware.
+    """
+
+    class ConfigRiak(ConfigDict):
```
hodgestar: Perhaps this could go into ``vumi.config``?

hodgestar:
Many :cake: :cake: :cake: for doing this!

vumi/middleware/logging.py
rudigiesler: Yes, I made `BaseMiddlewareConfig` after I made all these configs, so I'll change them to extend that instead.

vumi/middleware/tests/test_base.py
```diff
@@ -9,8 +11,16 @@
 from vumi.tests.helpers import VumiTestCase
 
 
-class ToyMiddleware(BaseMiddleware):
+class ToyMiddlewareConfig(Config):
```
hodgestar: Should this also inherit from ``BaseMiddlewareConfig``?

vumi/middleware/tests/test_base.py
```diff
@@ -280,9 +288,12 @@ def test_parse_yaml(self):
 
     @inlineCallbacks
     def test_sort_by_priority(self):
-        priority2 = ToyMiddleware('priority2', {"priority": 2}, self)
-        priority1_1 = ToyMiddleware('priority1_1', {"priority": 1}, self)
-        priority1_2 = ToyMiddleware('priority1_2', {"priority": 1}, self)
+        priority2 = ToyMiddleware('priority2', {}, self)
```
hodgestar: Why did specifying the priority in the the constructor stop working?

hodgestar:
Left two last comments, otherwise looks good.

vumi/middleware/tests/test_base.py
rudigiesler: The second parameter is the config dict. We used to just tag the parameter on as an extra key in the config dict, but we can't do this when using a config class (unless maybe we add it to the BaseMiddlewareConfig?).

At the moment I've made it a property of the middleware class (`mw.consume_priority`). I'm not sure if this is a better way of doing it, or if it should rather be added to the BaseMiddlwareConfig.

hodgestar: Ah! I see this is just for testing `_sort_by_priority`. Happy with it the way it is.

hodgestar:
:+1:

PR #912, opened 2015-02-16
Change session length middleware to not overwrite existing time metadata

justinvdm:
At the moment we overwrite existing time metadata.

### DIFF
--- a/vumi/middleware/session_length.py
+++ b/vumi/middleware/session_length.py
@@ -42,37 +42,43 @@ class SessionLengthMiddleware(BaseMiddleware):
     def teardown_middleware(self):
         yield self.redis.close_manager()
 
+    def _generate_redis_key(self, message, address):
+        return '%s:%s:%s' % (
+            message.get('transport_name'), address, 'session_created')
+
+    @inlineCallbacks
+    def _set_redis_time(self, redis_key, time):
+        if (yield self.redis.get(redis_key)) is None:
+            yield self.redis.setex(redis_key,  self.timeout, str(time))
+
     def _set_message_session_metadata(self, message, field, value):
-        if not message['helper_metadata'].get(self.field_name):
-            message['helper_metadata'][self.field_name] = {}
-        message['helper_metadata'][self.field_name][field] = value
+        metadata = message['helper_metadata'].setdefault(self.field_name, {})
+        metadata.setdefault(field, value)
 
     def _set_session_start_time(self, message, time):
+        time = float(time)
         self._set_message_session_metadata(message, 'session_start', time)
 
     def _set_session_end_time(self, message, time):
+        time = float(time)
         self._set_message_session_metadata(message, 'session_end', time)
 
     @inlineCallbacks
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review, I think.

vumi/middleware/session_length.py
```diff
@@ -42,37 +42,43 @@ def setup_middleware(self):
...
         yield self.redis.close_manager()
 
+    def _generate_redis_key(self, message, address):
+        return '%s:%s:%s' % (
+            message.get('transport_name'), address, 'session_created')
+
+    @inlineCallbacks
+    def _set_redis_time(self, redis_key, time):
+        if (yield self.redis.get(redis_key)) is None:
+            yield self.redis.setex(redis_key,  self.timeout, str(time))
```
rudigiesler: I'm not quite sure if our current redis library supports this, but the redis command `SET key value EX timeout NX` will do this in one, atomic command. http://redis.io/commands/set

rudigiesler:
Just the one comment, the rest looks good to me.

vumi/middleware/session_length.py
```diff
@@ -42,37 +42,43 @@ def setup_middleware(self):
...
     def _set_session_end_time(self, message, time):
+        time = float(time)
         self._set_message_session_metadata(message, 'session_end', time)
 
     @inlineCallbacks
     def _process_message(self, message, redis_key):
         if message.get('session_event') == self.SESSION_NEW:
             start_time = self.clock.seconds()
-            yield self.redis.setex(redis_key,  self.timeout, str(start_time))
+            yield self._set_redis_time(redis_key, start_time)
```
hodgestar: I think I was expecting some like:
```python
if message['helper_metadata'][self.fieldname].get('session_start') is not None:
    ...
```
rather than the Redis check? I think Redis should always be updated if we set something of the message because otherwise the two might get out of sync (apologies if my comments elsewhere suggested something different).

justinvdm: >  Justin van der Merwe [4:53 PM] 
>  hodgestar: should we change the middleware to not overwrite values but still overwrite the redis value, or should it overwrite neither?
>
>  Justin van der Merwe [4:54 PM]
>  overwriting neither seems sensible
>
>  Simon Cross [4:54 PM] 
>  +1

I guess you understood my suggestion as meaning "overwrite both the redis value and the message start time value only if the message has no start time value".

I'm wondering if we would ever need to explicitly set message start times somewhere else (without changing the redis value), but I guess we can wait for that use case to appear before trying to support that.

I can't think of a way that the two values could get out of sync, but I'm sure are ways I'm not thinking of. +1 on the suggested approach

justinvdm:
Ready for re-review.

hodgestar:
:+1: once we understand what the Travis failures are about (looks like they're unrelated).

hodgestar:
Builds restarted.

hodgestar:
:+1:

PR #913, opened 2015-02-19
Reduce default concurrency in message store resource

jerith:
Unless we have lots of cores available for the message store exports, the concurrency hurts us instead of helping us.


justinvdm:
:+1:

PR #914, opened 2015-02-19
SMPP transport tps throttling should handle reconnects

jerith:
Similar to #909, but for TPS-based throttling.


justinvdm:
:+1: when travis is happy.

It might be useful if our throttling-related log messages mentioned what kind of throttling they were talking about, but not sure how possible that is.

PR #915, opened 2015-02-19
Fix session length middleware redis config field

justinvdm:
At the moment, we use `redis` as the config field name, while tests helpers use `redis_manager`. This means that tests will get `{}` as the redis config, and use default redis settings instead.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #916, opened 2015-02-20
Format message store timestamp indexes more rigorously

hodgestar:
Currently they're formatted using ``str(msg['timestamp'])``.

See https://github.com/praekelt/vumi-go/issues/1201 for an example of the issues this causes.

Probably they should be formatted with the Vumi timestamp format.


hodgestar:
Ready for review (on Monday).

rudigiesler:
:+1:

PR #917, opened 2015-02-25
More permissive timestamp parsing

jerith:
`vumi.message.parse_vumi_date()` requires the date being parsed to have microseconds, but it's useful to parse dates without microseconds as well.

### DIFF
--- a/vumi/message.py
+++ b/vumi/message.py
@@ -11,6 +11,8 @@ from vumi.utils import to_kwargs
 
 # This is the date format we work with internally
 VUMI_DATE_FORMAT = "%Y-%m-%d %H:%M:%S.%f"
+# Same as above, but without microseconds (for more permissive parsing).
+_VUMI_DATE_FORMAT_NO_MICROSECONDS = "%Y-%m-%d %H:%M:%S"
 
 
 def format_vumi_date(timestamp):
@@ -36,7 +38,7 @@ def parse_vumi_date(value):
     """
     date_format = VUMI_DATE_FORMAT
     if "." not in value:
-        date_format = date_format.split(".")[0]
+        date_format = _VUMI_DATE_FORMAT_NO_MICROSECONDS
     return datetime.strptime(value, date_format)
 
 
--- a/vumi/message.py
+++ b/vumi/message.py
@@ -37,7 +37,9 @@ def parse_vumi_date(value):
         A datetime object representing the timestamp.
     """
     date_format = VUMI_DATE_FORMAT
-    if "." not in value:
+    # We only look at the last ten characters, because that's where the "."
+    # will be in a valid serialised timestamp with microseconds.
+    if "." not in value[-10:]:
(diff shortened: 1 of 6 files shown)


vumi/message.py
```diff
@@ -27,12 +27,17 @@ def format_vumi_date(timestamp):
...
+
     :param str value:
         The string to parse.
     :return datetime:
         A datetime object representing the timestamp.
     """
-    return datetime.strptime(value, VUMI_DATE_FORMAT)
+    date_format = VUMI_DATE_FORMAT
+    if "." not in value:
+        date_format = date_format.split(".")[0]
```
hodgestar: I'm not wild about this way of detecting milliseconds and splitting the format. Maybe a ``len(value) <= ...`` check would be better? And perhaps we can just have a (possibly `_` private) constant for `_VUMI_DATE_FORMAT_NO_MS`?

hodgestar:
Reviewed. Tx for doing the clean up. Left one comment.

It's perhaps worthwhile mentioning that `str(datetime)` results in pseudo-Vumi timestamps without µs. Are there any other cases where we're dropping µs for this that you had in mind?

jerith:
There's a specific vumi-go test failure on my machine that inspired this, but we also probably have a whole lot of stored data that uses both formats.

vumi/message.py
jerith: I think looking for a dot to detect the microseconds is better than using the length of the string, but I'm okay with a second (private) format constant.

hodgestar: My thought was more than a len check is much faster. If one attempts to
parse arbitrary values with this, which I think our JSON decoder does, the
value string could get quite long.

jerith: `"." not in value` is O(`len(value)`), which isn't really all that expensive. (We've probably already spent at least that much time parsing JSON containing that value, so if there's a DoS attack it's already succeeded by the time we get here.) If we want to do micro-optimisations like this, we should take measurements and check that there's some benefit to counter the loss in clarity.

hodgestar: Checking the length is `O(1)` which makes this not a micro-optimization?

I think I'm really just balking at the idea of checking all the values in our messages for `.` which feels to me like a rather unexpected thing to have happen as a result of parsing dates (of course having fields unexpectedly turned into datetimes wasn't great to begin with). I don't find the length check any less clear personally (it's just checking a different features to distinguish the two cases) but we could always add a comment if needed.

Anyway, this will be easy enough to change later so leaving it up to you.

hodgestar:
Left one final comment about the `.` issue, but leaving whether or not to address that up to you.

:+1:

vumi/message.py
jerith: We could check the last ten characters or so instead of the whole string if you'd prefer. A quick test on my system indicates that scanning a 50MB string for a "." is about an order of magnitude cheaper than parsing that 50MB string out of JSON, though.

hodgestar:
:+1:

PR #918, opened 2015-02-26
Decrement _in_progress counter when a consumer breaks

jerith:
The consumer in ``vumi.service`` maintains an ``_in_progress`` counter to track how many messages are currently being handled. If processing a message raises an exception, that will prevent the worker with the broken consumer from shutting down cleaning because it'll keep waiting for ``_in_progress`` to reach zero.

### DIFF
--- a/vumi/service.py
+++ b/vumi/service.py
@@ -333,11 +333,16 @@ class Consumer(object):
     @inlineCallbacks
     def consume(self, message):
         self._in_progress += 1
-        result = yield self.consume_message(self.message_class.from_json(
-                                            message.content.body))
-        self._in_progress -= 1
-        if self._testing:
-            self.channel.message_processed()
+        try:
+            result = yield self.consume_message(
+                self.message_class.from_json(message.content.body))
+        finally:
+            # If we get an exception here the consumer's already pretty much
+            # broken, but we still decrement the _in_progress counter so we
+            # don't wait forever for it during shutdown.
+            self._in_progress -= 1
+            if self._testing:
+                self.channel.message_processed()
         if result is not False:
             yield self.channel.basic_ack(message.delivery_tag, False)
         else:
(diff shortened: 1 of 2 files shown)


vumi/tests/test_service.py
```diff
@@ -57,6 +57,39 @@ def test_consume_with_prefetch(self):
...
+            'vumi', 'test.routing.key', message.content)
+        # Wait for the processing to start and assert that it's in progress.
+        yield start_d
+        self.assertEqual(consumer._in_progress, 1)
+        # Unpause the processing and wait for it to finish.
+        pause_d.callback(None)
+        yield self.worker_helper.kick_delivery()
+        # If we get here without timing out, everything should be happy.
+        self.assertEqual(consumer._in_progress, 0)
+        self.flushLoggedErrors()
```
hodgestar: We should assert the expected exception was flushed. Might be easiest to just make a local ``DummyException(Exception)`` class to check for.

vumi/service.py
```diff
@@ -331,11 +333,16 @@ def _check_notify(self):
...
+        try:
+            result = yield self.consume_message(
+                self.message_class.from_json(message.content.body))
+        finally:
+            # If we get an exception here the consumer's already pretty much
+            # broken, but we still decrement the _in_progress counter so we
+            # don't wait forever for it during shutdown.
+            self._in_progress -= 1
+            if self._testing:
+                self.channel.message_processed()
```
hodgestar: Just double-checking that the ``if self._testing: ...`` block should be inside the ``finally:`` (i.e. does an exception count as having processed the message for testing purposes). I think the answer is yes, but I'm not entirely sure.

hodgestar:
Two small comments. Otherwise looks good.

vumi/service.py
jerith: Yes. The only thing that's used for is telling the fake broker that we're done with a message so we can wait for the delivery run to finish in tests, and we want that to continue working even if the consumer is broken. (We need it for the test in this PR, if nothing else.)

vumi/tests/test_service.py
jerith: Done.

hodgestar:
:+1: as soon as Travis is happy.

PR #919, opened 2015-02-26
Allow the session length middleware's redis key namespace to be configurable

justinvdm:
At the moment, the session length middleware uses the message's `transport_name` to namespace the address-timestamp key-value pairs in redis. When we put the session length middleware on load balanced transports, the redis key used for start-of-session inbound messages is the actual transport name, not the exposed transport name rewritten by the load balancer (since the middleware sees the messages before the load balancer does).

One way around this is to allow the middleware's namespace to be configurable, and default to the message's `transport_name`.


justinvdm:
Ready for review.

hodgestar:
:+1: when Travis is happy.

The build failures looked unrelated so I restarted the two builds that failed.

PR #921, opened 2015-02-27
Better handling of Redis connections in tests

jerith:
If we have multiple Redis connections in a test, we sometimes get errors during cleanup.


calston:
:+1:

PR #922, opened 2015-02-27
Session length middleware breaks when no transport name is given

justinvdm:
We try generate redis keys using the transport name, this breaks when the message has no transport name as of #919.


justinvdm:
Ready for review.

justinvdm:
Fixed by #923.

PR #900, opened 2015-03-02
Test FakeRedis against real Redis

jerith:
There are a number of differences between the behaviours of FakeRedis and a real Redis client+server combination. We really should run the same set of tests for both and fix the differences.

### DIFF
--- a/vumi/persist/fake_redis.py
+++ b/vumi/persist/fake_redis.py
@@ -26,6 +26,13 @@ def call_to_deferred(deferred, func, *args, **kw):
     execute(func, *args, **kw).chainDeferred(deferred)
 
 
+class ResponseError(Exception):
+    """
+    Exception class for things we throw to match the real Redis client
+    libraries.
+    """
+
+
 class FakeRedis(object):
     """In process and memory implementation of redis-like data store.
 
@@ -171,7 +178,7 @@ class FakeRedis(object):
         else:
             cursor = str(start + i + 1)
 
-        return cursor, fnmatch.filter(output, match)
+        return [cursor, fnmatch.filter(output, match)]
 
     @maybe_async
     def flushdb(self):
@@ -188,6 +195,7 @@ class FakeRedis(object):
     def set(self, key, value):
         value = self._encode(value)  # set() sets string value
         self._set_key(key, value)
+        return True
 
     @maybe_async
     def setex(self, key, time, value):
@@ -387,7 +395,7 @@ class FakeRedis(object):
 
     @maybe_async
     def zcount(self, key, min, max):
-        return str(len(self.zrangebyscore.sync(self, key, min, max)))
+        return len(self.zrangebyscore.sync(self, key, min, max))
 
     @maybe_async
     def zscore(self, key, value):
@@ -416,12 +424,13 @@ class FakeRedis(object):
 
     @maybe_async
     def lpush(self, key, obj):
-        self._setdefault_key(key, []).insert(0, obj)
+        self._setdefault_key(key, []).insert(0, self._encode(obj))
+        return self.llen.sync(self, key)
 
     @maybe_async
     def rpush(self, key, obj):
-        self._setdefault_key(key, []).append(obj)
-        return self.llen.sync(self, key) - 1
+        self._setdefault_key(key, []).append(self._encode(obj))
+        return self.llen.sync(self, key)
 
     @maybe_async
     def lrange(self, key, start, end):
@@ -435,6 +444,7 @@ class FakeRedis(object):
     @maybe_async
     def lrem(self, key, value, num=0):
         removed = [0]
+        value = self._encode(value)
 
         def keep(v):
             if v == value and (num == 0 or removed[0] < abs(num)):
@@ -468,6 +478,7 @@ class FakeRedis(object):
             # want to keep.
             del lval[stop + 1:]
         del lval[:start]
+        return True
 
     # Expiry operations
 
@@ -484,7 +495,7 @@ class FakeRedis(object):
     def ttl(self, key):
         delayed = self._expiries.get(key)
         if delayed is not None and delayed.active():
-            return int(delayed.getTime() - self.clock.seconds())
+            return round(delayed.getTime() - self.clock.seconds())
         return None
 
     @maybe_async
@@ -508,9 +519,15 @@ class Zset(object):
             end = None
         return start, end
 
+    def _to_float(self, value):
+        try:
+            return float(value)
+        except (ValueError, TypeError):
+            raise ResponseError("value is not a valid float")
+
     def zadd(self, **valscores):
         new_zval = [val for val in self._zval if val[1] not in valscores]
-        new_zval.extend((float(score), value)
+        new_zval.extend((self._to_float(score), value)
                         for value, score in valscores.items())
         new_zval.sort()
         added = len(new_zval) - len(self._zval)
--- a/vumi/persist/fake_redis.py
+++ b/vumi/persist/fake_redis.py
@@ -293,9 +293,16 @@ class FakeRedis(object):
 
     @maybe_async
     def hincrby(self, key, field, amount=1):
-        value = self._data.get(key, {}).get(field, "0")
+        try:
+            value = self._data.get(key, {}).get(field, "0")
+        except AttributeError:
+            raise ResponseError("WRONGTYPE Operation against a key holding"
+                                " the wrong kind of value")
         # the int(str(..)) coerces amount to an int but rejects floats
-        value = int(value) + int(str(amount))
+        try:
+            value = int(value) + int(str(amount))
+        except (TypeError, ValueError):
+            raise ResponseError("value is not an integer or out of range")
         self._setdefault_key(key, {})[field] = str(value)
         return value
 
--- a/vumi/persist/fake_redis.py
+++ b/vumi/persist/fake_redis.py
@@ -272,8 +272,8 @@ class FakeRedis(object):
 
     @maybe_async
     def hgetall(self, key):
-        return dict((self._encode(k), self._encode(v)) for k, v in
-            self._data.get(key, {}).items())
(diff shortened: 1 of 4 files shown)


jerith:
Whoops, I left some tests commented. I'll fix that shortly.

vumi/persist/tests/test_fake_redis.py
```diff
@@ -1,491 +1,747 @@
...
+
+    def wait(self, redis, delay):
+        redis._fake_redis.clock.advance(delay)
+        d = Deferred()
+        reactor.callLater(delay, d.callback, None)
+        return d
+
+
+# TestFakeRedisVerify.skip = "foo"
+# TestFakeRedisVerifyAsync.skip = "foo"
```
justinvdm: Should these be here?

jerith: Good catch, thanks. I'll remove them.

justinvdm:
Would it be worth running the vumi-go tests using the fake redis changes? Not sure if that needs to be done before this PR lands though.

jerith:
> Would it be worth running the vumi-go tests using the fake redis changes? Not sure if that needs to be done before this PR lands though.

Good idea, I'll do that now. Thanks!

jerith:
vumi-go is happy with the new fake Redis. (Most of the changes were response values on write operations, so we were pretty much ignoring them anyway.)

justinvdm:
> vumi-go is happy with the new fake Redis. (Most of the changes were response values on write operations, so we were pretty much ignoring them anyway.)

Awesome, thanks for checking.

vumi/persist/tests/test_fake_redis.py
```diff
@@ -1,491 +1,750 @@
...
+    Each test in here should explain why verification isn't possible.
+    """
 
     @inlineCallbacks
     def test_scan_simple(self):
+        """
+        Scanning returns keys in an order that depends on arbitrary state in
+        the Redis server, so we can't fake it in a way that's identical to real
+        Redis.
+        """
```
justinvdm: I would've thought testing this against both fake and real redis is possible if we just sort the returned keys before asserting on them (along with a docstring explaining why sorting is necessary in the test). Any reason why that wouldn't work that I'm missing?

justinvdm: Oh, right, we can't assume that a set of keys returned by a scan call will always be the same set of keys. For some reason I assumed that would be the case, nevermind.

vumi/persist/tests/test_fake_redis.py
```diff
@@ -1,491 +1,750 @@
...
+                errors.append(e)
+            else:
+                results.append(result)
+
+        # First, handle errors.
+        if errors:
+            if results:
+                self._test_case.fail(
+                    "One redis returned %r and the other raised %r" % (
+                        results[0], errors[0]))
```
justinvdm: It would be helpful to know which one returned something and which one raised something, might save us some confusion.

vumi/persist/tests/test_fake_redis.py
```diff
@@ -1,491 +1,750 @@
...
+    def get_redis(self, **kwargs):
+        from vumi.persist.txredis_manager import TxRedisManager
+        # Fake redis
+        fake_redis = FakeRedis(async=True, **kwargs)
+        self.add_cleanup(fake_redis.teardown)
+        # Real redis
+        config = {
+            'FAKE_REDIS': 'yes',
+            'key_prefix': 'redistest',
+        }
```
justinvdm: If I'm understanding this correctly setting `FAKE_REDIS` to `'yes'` would cause us to create a fake redis instance, even though this config is for real redis. Any reason why we do that here?

vumi/persist/tests/test_fake_redis.py
```diff
@@ -1,491 +1,750 @@
...
+                raise errors[0]
+
+        # Now handle results.
+        self._test_case.assertEqual(
+            results[0], results[1],
+            "Fake redis (a) and real redis (b) responses different:")
+        returnValue(results[0])
+
+    def __getattr__(self, name):
+        return lambda *args, **kw: self._perform_operation(name, *args, **kw)
```
justinvdm: I think we need tests for `FakeRedisWrapper`

justinvdm: err, s/`FakeRedisWrapper`/`RedisPairWrapper`

vumi/persist/txredis_manager.py
```diff
@@ -154,6 +174,15 @@ def ttl(self, key):
         d.addCallback(lambda r: (None if r < 0 else r))
         return d
 
+    def persist(self, key):
+        self._send('PERSIST', key)
+        return self.getResponse()
+
+    def type(self, key):
+        d = self.get_type(key)
+        d.addCallback(lambda r: r if r is not None else 'none')
```
justinvdm: Any reason why we turn `None`s into `'none'`s? To ensure all results are string?

vumi/persist/txredis_manager.py
```diff
@@ -154,6 +174,15 @@ def ttl(self, key):
         d.addCallback(lambda r: (None if r < 0 else r))
         return d
 
+    def persist(self, key):
+        self._send('PERSIST', key)
+        return self.getResponse()
```
justinvdm: Does txredis not implement this? If not, could we add a docstring mentioning that like we do above?

vumi/persist/txredis_manager.py
justinvdm: Ah, is `'none'` maybe what the sync client returns?

justinvdm:
I might just be not finding it in the diff, I don't think we have tests for `set` (including making sure that we turn `OK`s into `True`s). Could we add test(s) for it?

justinvdm:
Looks good, left some comments.

vumi/persist/tests/test_fake_redis.py
jerith: Yeah. I tried to rewrite the tests in a way that would let us walk the scan over the whole set of keys and check that we got the same set of results at the end, but that ended up being more effort than it's really worth.

vumi/persist/tests/test_fake_redis.py
jerith: I thought about that, but my vague ideas made the code rather messier. I'll see if I can come up with a better mechanism.

vumi/persist/tests/test_fake_redis.py
jerith: If the `VUMITEST_REDIS_DB` environment variable is set, this is overridden. At the moment, setting `FAKE_REDIS` doesn't change any behaviour, but that may change in the future. (For example, it might be useful to ensure that all Redis managers created in tests have it set, although I can't think of a practical way to do that at the moment.)

vumi/persist/tests/test_fake_redis.py
jerith: I thought about it, but couldn't come up with a good way to test it that didn't involve test code of a similar level of complexity, which would then have to be tested itself, etc. I'll see if I can find a better way.

vumi/persist/txredis_manager.py
jerith: Yes. There's code in txredis that turns `'none'` into `None`, but leaves all the other return values of the `TYPE` command as strings. I'll add a comment.

vumi/persist/txredis_manager.py
jerith: Correct. I'll do that, thanks.

jerith:
> I might just be not finding it in the diff, I don't think we have tests for `set` (including making sure that we turn `OK`s into `True`s). Could we add test(s) for it?

I'd prefer to leave that for another branch -- the purpose of this one is to build the verification infrastructure. I only added new tests for things that changed in ways that required modifications to other tests, like `lpush` and `rpush`. There are probably a bunch of other methods that aren't being tested explicitly (or perhaps even at all) and I don't want this branch to get too big.

vumi/persist/tests/test_fake_redis.py
justinvdm: Ok cool, thanks. Happy for it to be a separate PR.

vumi/persist/txredis_manager.py
justinvdm: +1, thanks

justinvdm:
> I'd prefer to leave that for another branch -- the purpose of this one is to build the verification infrastructure. I only added new tests for things that changed in ways that required modifications to other tests, like lpush and rpush. There are probably a bunch of other methods that aren't being tested explicitly (or perhaps even at all) and I don't want this branch to get too big.

Happy for that to happen in a separate PR

vumi/persist/fake_redis.py
```diff
@@ -285,9 +293,16 @@ def hvals(self, key):
...
+            value = self._data.get(key, {}).get(field, "0")
+        except AttributeError:
+            raise ResponseError("WRONGTYPE Operation against a key holding"
+                                " the wrong kind of value")
         # the int(str(..)) coerces amount to an int but rejects floats
-        value = int(value) + int(str(amount))
+        try:
+            value = int(value) + int(str(amount))
+        except (TypeError, ValueError):
+            raise ResponseError("value is not an integer or out of range")
```
justinvdm: Do we have tests for these cases:
  - using `"0"` as the default field
  - `hincrby` against a key holding the wrong kind of value
  - `hincrby` with a non-integer or out of range value

If not, could we add some?

jerith: Theses cases are all covered in `test_hincrby`.

justinvdm: Thanks, missed those.

justinvdm:
:+1: when travis is happy.

Should we make tickets for `RedisPairWrapper` and `set` tests?

jerith:
I opened #926 for the test coverage improvements.

PR #929, opened 2015-03-10
It's easy to fetch all search results accidentally with real_search.

hodgestar:
Currently passing ``start=None`` to ``real_search`` causes it to ignore ``rows`` and just return all search results (which will often time out).

### DIFF
--- a/vumi/persist/riak_manager.py
+++ b/vumi/persist/riak_manager.py
@@ -293,16 +293,10 @@ class RiakManager(Manager):
 
     def real_search(self, modelcls, query, rows=None, start=None):
         rows = 1000 if rows is None else rows
+        start = 0 if start is None else start
         bucket_name = self.bucket_name(modelcls)
         bucket = self.client.bucket(bucket_name)
-        if start is not None:
-            return self._search_iteration(bucket, query, rows, start)
-        keys = []
-        new_keys = self._search_iteration(bucket, query, rows, 0)
-        while new_keys:
-            keys.extend(new_keys)
-            new_keys = self._search_iteration(bucket, query, rows, len(keys))
-        return keys
+        return self._search_iteration(bucket, query, rows, start)
 
     def riak_enable_search(self, modelcls):
         bucket_name = self.bucket_name(modelcls)
--- a/vumi/persist/model.py
+++ b/vumi/persist/model.py
@@ -474,6 +474,19 @@ class Model(object):
         """
         Performs a real riak search, does no inspection on the given query.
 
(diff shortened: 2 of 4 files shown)


vumi/persist/riak_manager.py
```diff
@@ -293,16 +293,10 @@ def _search_iteration(self, bucket, query, rows, start):
 
     def real_search(self, modelcls, query, rows=None, start=None):
         rows = 1000 if rows is None else rows
+        start = 0 if start is None else start
```
jerith: I'd prefer to throw an exception if we get `start=None` so we can find places that are using this incorrectly and fix them instead of having them silently return fewer results than they did before.

hodgestar: I think raising an exception for the default value is a bit crazy? Everywhere I can find that uses real_search currently passes in a start value and expects it to be honoured.

jerith: If this were a new method, I'd completely agree. What we're doing here is changing the behaviour of the default case, and throwing an error for the now-unsupported behaviour (in any places it's used that we haven't found) is better than silently returning fewer results than we had before. We'd have to document it clearly, of course -- probably in the exception.

(If `start` were before `rows` in the parameter list, I'd suggest removing the default value and making it mandatory.)

hodgestar: I don't think these other places exist? And if there are a couple of places we can fix them when we find them?

The situation in which the current implementation returns more results is a bit corner case anyway since it requires there to be more results than rows (so more than 1000) by default, but few enough to return in a reasonable time frame.

The cost of raising the exception is having yet another ugly function in the code base that we have to clean up later.

jerith: > I don't think these other places exist? And if there are a couple of places we can fix them when we find them?

How will we find them if we don't raise an exception? 

> The cost of raising the exception is having yet another ugly function in the code base that we have to clean up later.

The cost of not raising the exception is potentially introducing subtle data corruption bugs into existing code. Having been on the receiving end of that kind of change more than once, I'll take the ugly function every time.

jerith:
Anyway, I'll leave this decision to you. :+1: other than that.

hodgestar:
Tx. While this PR was open @rudigiesler fixed the contacts api in https://github.com/praekelt/go-contacts-api/commit/cd9b78f3121fbafcff34d7420f5140a5cb0b3339 but it doesn't appear to have helped in practice. I'd like to understand better what's going on before landing this (in case something more needs to change).

PR #930, opened 2015-03-18
Allow setting session_event when sending non-reply messages.

hodgestar:
Currently the session_event is hard coded to SESSION_NONE, which makes sense in many cases. However, MT session transports require the initial send to have SESSION_NEW.


hodgestar:
Ready for review.

rudigiesler:
:+1:

PR #931, opened 2015-03-19
Add support for PFADD and PFCOUNT Redis commands

jerith:
This needs to include both Redis clients and FakeRedis.

### DIFF
--- a/vumi/persist/fake_redis.py
+++ b/vumi/persist/fake_redis.py
@@ -6,6 +6,7 @@ from itertools import takewhile, dropwhile
 import os
 from zlib import crc32
 
+from hyperloglog import HyperLogLog
 from twisted.internet import reactor
 from twisted.internet.defer import Deferred, execute
 from twisted.internet.task import Clock
@@ -512,6 +513,21 @@ class FakeRedis(object):
             delayed.cancel()
             return 1
         return 0
+
+    # HyperLogLog operations
+
+    @maybe_async
+    def pfadd(self, key, *values):
+        hll = self._setdefault_key(key, HyperLogLog(0.01))
+        old_card = hll.card()
+        for value in values:
+            hll.add(value)
+        return hll.card() != old_card
+
+    @maybe_async
+    def pfcount(self, key):
+        hll = self._data.get(key, HyperLogLog(0.01))
+        return len(hll)
 
 
 class Zset(object):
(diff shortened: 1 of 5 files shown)


vumi/persist/fake_redis.py
```diff
@@ -513,6 +514,21 @@ def persist(self, key):
...
+        hll = self._setdefault_key(key, HyperLogLog(0.01))
+        old_card = hll.card()
+        for value in values:
+            hll.add(value)
+        return hll.card() != old_card
+
+    @maybe_async
+    def pfcount(self, key):
+        hll = self._data.get(key, HyperLogLog(0.01))
+        return len(hll)
```
hodgestar: I was expecting ``hll.card()`` here instead of ``len(hll)``. Are they the same?

vumi/persist/fake_redis.py
```diff
@@ -513,6 +514,21 @@ def persist(self, key):
             return 1
         return 0
 
+    # HyperLogLog operations
+
+    @maybe_async
+    def pfadd(self, key, *values):
+        hll = self._setdefault_key(key, HyperLogLog(0.01))
```
hodgestar: Redis' quoted accuracy is `0.81%`. I'm guessing `0.01` is `1%`? Does that mean we should be using `0.0081` or `0.01`?

hodgestar:
Left a few minor comments but generally looks awesome.

I think we could implement `pfmerge` fairly easily using `hll.update()` but that's like something for another PR.

justinvdm:
Looks good to me too, got nothing to add that hodgestar hasn't already mentioned.

vumi/persist/fake_redis.py
jerith: `hll.card()` returns a float. `len(hll)` is equivalent to `round(hll.card())` and is what the usage section of hyperloglog's README uses.

vumi/persist/fake_redis.py
jerith: Both Redis and `HyperLogLog(0.01)` use 14 bits from the hash function, which gives an error of approximately 0.81%. `HyperLogLog(0.0081)` is just over the boundary between 14 and 15 bits, so it actually ends up with a different (smaller) error. (`0.0082` uses 14 bits, but it's uglier than `0.01`.)

hodgestar:
:+1:

PR #933, opened 2015-03-23
Unique address count approximations

justinvdm:
The plan is to use Redis's HyperLogLog functionality for this.

### DIFF
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -253,7 +253,7 @@ class MessageStore(object):
         index_page = yield self.batch_inbound_keys_with_addresses(batch_id)
         while index_page is not None:
             for key, timestamp, addr in index_page:
-                self.cache.add_from_addr(batch_id, addr, timestamp)
+                self.cache.add_from_addr(batch_id, addr)
                 old_key = key_manager.add_key(key, timestamp)
                 if old_key is not None:
                     key_count += 1
@@ -280,7 +280,7 @@ class MessageStore(object):
         index_page = yield self.batch_outbound_keys_with_addresses(batch_id)
         while index_page is not None:
             for key, timestamp, addr in index_page:
-                self.cache.add_to_addr(batch_id, addr, timestamp)
+                self.cache.add_to_addr(batch_id, addr)
                 old_key = key_manager.add_key(key, timestamp)
                 if old_key is not None:
                     key_count += 1
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -250,9 +250,10 @@ class MessageStore(object):
             start_timestamp, self.cache.TRUNCATE_MESSAGE_KEY_COUNT_AT)
         key_count = 0
 
-        index_page = yield self.batch_inbound_keys_with_timestamps(batch_id)
+        index_page = yield self.batch_inbound_keys_with_addresses(batch_id)
         while index_page is not None:
-            for key, timestamp in index_page:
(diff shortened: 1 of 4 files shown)


justinvdm:
Not yet ready for review, still need to update the cache reconciliation stuff.

vumi/components/message_store_cache.py
```diff
@@ -389,26 +385,21 @@ def get_from_addrs(self, batch_id, asc=False):
...
+        http://redis.io/commands/pfcount
         """
-        # NOTE: Disabled because this doesn't scale to large batches.
-        #       See https://github.com/praekelt/vumi/issues/877
-
-        # return self.redis.zcard(self.from_addr_key(batch_id))
-        return 0
+        return self.redis.pfcount(self.from_addr_key(batch_id))
 
     def add_to_addr(self, batch_id, to_addr, timestamp):
```
justinvdm: We no longer use `timestamp`, but I'm wondering if we should keep it around for backwards compatibility, and in case we end up using it later.

justinvdm:
Ready for review.

vumi/components/message_store_cache.py
jerith: No, we can remove it. There shouldn't be anything outside the message store using this.

vumi/components/tests/test_message_store.py
```diff
@@ -1249,6 +1250,12 @@ def test_reconcile_cache(self):
         self.assertFalse((yield self.store.needs_reconciliation(batch_id)))
         self.assertFalse(
             (yield self.store.needs_reconciliation(batch_id, delta=0)))
+
+        inbound_uniques = yield self.store.cache.count_from_addrs(batch_id)
+        self.assertEqual(inbound_uniques, 1)
+        outbound_uniques = yield self.store.cache.count_to_addrs(batch_id)
+        self.assertEqual(outbound_uniques, 1)
```
jerith: These tests should probably be updated to use more than one address, and ideally different numbers of addresses (and messages, really) between inbound and outbound to make sure we're updating the correct keys with the correct values.

vumi/components/message_store_cache.py
justinvdm: Cool, will do.

vumi/components/tests/test_message_store.py
justinvdm: Makes me wonder why the current code isn't using different numbers of messages in the first place, but will do.

jerith: Probably because I didn't think about it when I was writing the tests. (The new message store tests do it that way, though.)

justinvdm:
Ready for re-review.

vumi/components/message_store.py
```diff
@@ -250,9 +250,10 @@ def reconcile_inbound_cache(self, batch_id, start_timestamp):
             start_timestamp, self.cache.TRUNCATE_MESSAGE_KEY_COUNT_AT)
         key_count = 0
 
-        index_page = yield self.batch_inbound_keys_with_timestamps(batch_id)
+        index_page = yield self.batch_inbound_keys_with_addresses(batch_id)
         while index_page is not None:
-            for key, timestamp in index_page:
+            for key, timestamp, addr in index_page:
+                self.cache.add_from_addr(batch_id, addr)
```
jerith: Don't we need to `yield` here?

justinvdm:
Ready for another review.

jerith:
:+1:

PR #937, opened 2015-04-02
Reverse timestamp batch indexes for message store

jerith:
We often want message store index results in reverse time order, which currently means walking the entire index to get the results we need.

(I was originally planning to add similar indexes to stored events as well, but that changes the message store API and is therefore best left for a new PR.)

### DIFF
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -3,6 +3,7 @@
 
 """Message store."""
 
+from calendar import timegm
 from collections import defaultdict
 from datetime import datetime
 from uuid import uuid4
@@ -21,6 +22,29 @@ from vumi import log
 from vumi.components.message_store_cache import MessageStoreCache
 from vumi.components.message_store_migrators import (
     EventMigrator, InboundMessageMigrator, OutboundMessageMigrator)
+
+
+def to_reverse_timestamp(vumi_timestamp):
+    """
+    Turn a vumi_date-formatted string into a string that sorts in reverse order
+    and can be turned back into a timestamp later.
+
+    This is done by converting to a unix timestamp and subtracting it from
+    0xffffffffff (2**40 - 1) to get a number well outside the range
+    representable by the datetime module. The result is returned as a
+    hexadecimal string.
+    """
+    timestamp = timegm(parse_vumi_date(vumi_timestamp).timetuple())
+    return "%X" % (0xffffffffff - timestamp)
+
+
+def from_reverse_timestamp(reverse_timestamp):
+    """
+    Turn a reverse timestamp string (from `to_reverse_timestamp()`) into a
+    vumi_date-formatted string.
+    """
+    timestamp = 0xffffffffff - int(reverse_timestamp, 16)
+    return format_vumi_date(datetime.utcfromtimestamp(timestamp))
 
 
 class Batch(Model):
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -1100,15 +1100,5 @@ def key_with_ts_and_value_formatter(batch_id, result):
 
 
 def key_with_rts_and_value_formatter(batch_id, result):
-    value, key = result
-    prefix = batch_id + "$"
-    if not value.startswith(prefix):
-        raise ValueError(
-            "Index value %r does not begin with expected prefix %r." % (
-                value, prefix))
-    suffix = value[len(prefix):]
-    timestamp, delimiter, address = suffix.partition("$")
-    if delimiter != "$":
-        raise ValueError(
-            "Index value %r does not match expected format." % (value,))
-    return (key, from_reverse_timestamp(timestamp), address)
+    key, reverse_ts, value = key_with_ts_and_value_formatter(batch_id, result)
+    return (key, from_reverse_timestamp(reverse_ts), value)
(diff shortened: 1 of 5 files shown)


vumi/components/message_store.py
```diff
@@ -23,6 +24,29 @@
...
+    Turn a vumi_date-formatted string into a string that sorts in reverse order
+    and can be turned back into a timestamp later.
+
+    This is done by converting to a unix timestamp and subtracting it from
+    0xffffffffff (2**40 - 1) to get a number well outside the range
+    representable by the datetime module. The result is returned as a
+    hexadecimal string.
+    """
+    timestamp = timegm(parse_vumi_date(vumi_timestamp).timetuple())
+    return "%X" % (0xffffffffff - timestamp)
```
rudigiesler: Is there a reason why we're not using `2^^31 - 1` or `2^^63 - 1` (for 32 and 64 signed ints)? Although 32 bits will run out relatively soon.

rudigiesler: Ah, I just thought of negative timestamps for times before the epoch. Can we not use negative numbers?

jerith: The index values are sorted lexicographically, not numerically, so `-1` would sort before `-2`. I chose `2**40 - 1` because it's an integer number of octets and gives us several thousand years of headroom. Adding more bytes just makes the index values longer without adding any value.

justinvdm: Why do we convert to hex?

jerith: No particular reason, aside from a slightly more compact representation and a vague bias toward treating decimal values as increasing over time.

vumi/components/message_store.py
```diff
@@ -124,18 +154,24 @@ class InboundMessage(Model):
...
-            batches_with_timestamps.append(u"%s$%s" % (batch_id, timestamp))
-            batches_with_addresses.append(
+            self.batches_with_timestamps.append(
+                u"%s$%s" % (batch_id, timestamp))
+            self.batches_with_addresses.append(
                 u"%s$%s$%s" % (batch_id, timestamp, self.msg['from_addr']))
-        self.batches_with_timestamps = batches_with_timestamps
-        self.batches_with_addresses = batches_with_addresses
+            self.batches_with_addresses_reverse.append(
+                u"%s$%s$%s" % (batch_id, reverse_ts, self.msg['from_addr']))
```
justinvdm: Probably not worth doing much work on here since we are going to be using github.com/praekelt/vumi-message-store some time in the future, but I'm seeing quite a bit of repitition between `InboundMessage.save` and `OutboundMessage.save`. Would it maybe be worth removing that repitition in vumi-message-store, if these models were copied across as is?

jerith: I'd rather build support for this sort of thing into the base model/field code and get it out of the save method entirely. I was actually looking into doing that already.

justinvdm: Sounds good.

vumi/components/message_store.py
```diff
@@ -1000,18 +1071,44 @@ def has_next_page(self):
...
+    if not value.startswith(prefix):
+        raise ValueError(
+            "Index value %r does not begin with expected prefix %r." % (
+                value, prefix))
+    suffix = value[len(prefix):]
+    timestamp, delimiter, address = suffix.partition("$")
+    if delimiter != "$":
+        raise ValueError(
+            "Index value %r does not match expected format." % (value,))
+    return (key, from_reverse_timestamp(timestamp), address)
```
justinvdm: This looks identitical to `key_with_ts_and_value_formatter`, apart from the `from_reverse_timestamp()` call at the bottom. Could we not have this function just use `key_with_ts_and_value_formatter`, then then use `from_reverse_timestamp()` on the timestamp in the returned tuple?

rudigiesler:
:+1: from me, although it's probably worth letting justinvdm finish looking through it.

vumi/components/message_store.py
jerith: Excellent suggestion, sir. Thanks!

vumi/components/message_store.py
```diff
@@ -76,18 +100,24 @@ class OutboundMessage(Model):
...
         # We override this method to set our index fields before saving.
-        batches_with_timestamps = []
-        batches_with_addresses = []
-        timestamp = format_vumi_date(self.msg['timestamp'])
+        self.batches_with_timestamps = []
+        self.batches_with_addresses = []
+        self.batches_with_addresses_reverse = []
+        timestamp = self.msg['timestamp']
+        if not isinstance(timestamp, basestring):
+            timestamp = format_vumi_date(timestamp)
```
justinvdm: Do we have any test cases for checking that we leave string timestamps as is?

justinvdm:
One more minor comments, but looks good :) :+1:

vumi/components/message_store.py
jerith: The timestamp is currently always a `datetime` here, but there are enough other places where it could be one or the other that I put the check around all the calls to `parse_vumi_date()` and `format_vumi_date()`.

justinvdm: I see. Any reason why we shouldn't test this?

jerith: I don't think there's any particular reason to. If anything ever needs the other branch of the conditional, tests for that thing will cover it. I'd rather not add the additional runtime to the test suite unless there's enough value to justify it.

justinvdm: Not convinced, but seems unimportant enough that its not worth debating. Ok with us leaving it as is.

PR #938, opened 2015-04-02
Batch indexes for stored events

jerith:
Currently, stored events don't have any batch information associated with them. This means any batch operations involving events (such as cache rebuilds) require index lookups per outbound message, which cause load and performance issues.

(This depends on #937.)

### DIFF
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -446,6 +446,8 @@ class MessageStore(object):
             msg_record = yield self.outbound_messages.load(msg_id)
             if msg_record is not None:
                 batch_ids = msg_record.batches.keys()
+            else:
+                batch_ids = []
 
         for batch_id in batch_ids:
             event_record.batches.add_key(batch_id)
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -432,7 +432,7 @@ class MessageStore(object):
         returnValue(msg.msg if msg is not None else None)
 
     @Manager.calls_manager
-    def add_event(self, event):
+    def add_event(self, event, batch_ids=None):
         event_id = event['event_id']
         msg_id = event['user_message_id']
         event_record = yield self.events.load(event_id)
@@ -440,12 +440,18 @@ class MessageStore(object):
             event_record = self.events(event_id, event=event, message=msg_id)
         else:
             event_record.event = event
-        yield event_record.save()
 
-        msg_record = yield self.outbound_messages.load(msg_id)
-        if msg_record is not None:
-            for batch_id in msg_record.batches.keys():
-                yield self.cache.add_event(batch_id, event)
+        # If we aren't given batch_ids, get them from the outbound message.
+        if batch_ids is None:
+            msg_record = yield self.outbound_messages.load(msg_id)
+            if msg_record is not None:
+                batch_ids = msg_record.batches.keys()
+
+        for batch_id in batch_ids:
+            event_record.batches.add_key(batch_id)
+            yield self.cache.add_event(batch_id, event)
+
+        yield event_record.save()
 
     @Manager.calls_manager
     def get_event(self, event_id):
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -10,7 +10,7 @@ from uuid import uuid4
 import itertools
 import warnings
 
-from twisted.internet.defer import returnValue
+from twisted.internet.defer import inlineCallbacks, returnValue
 
 from vumi.message import (
     TransportEvent, TransportUserMessage, parse_vumi_date, format_vumi_date)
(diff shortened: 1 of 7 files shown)


vumi/components/message_store.py
```diff
@@ -431,12 +440,18 @@ def add_event(self, event):
...
-        yield event_record.save()
 
-        msg_record = yield self.outbound_messages.load(msg_id)
-        if msg_record is not None:
-            for batch_id in msg_record.batches.keys():
-                yield self.cache.add_event(batch_id, event)
+        # If we aren't given batch_ids, get them from the outbound message.
+        if batch_ids is None:
+            msg_record = yield self.outbound_messages.load(msg_id)
+            if msg_record is not None:
```
justinvdm: In what cases would we not have an outbound message for the event?

vumi/components/message_store.py
```diff
@@ -431,12 +440,18 @@ def add_event(self, event):
...
+                batch_ids = msg_record.batches.keys()
+
+        for batch_id in batch_ids:
```
justinvdm: If `msg_record` is `None` above, wouldn't `batch_ids` still be `None` at this point?

vumi/components/message_store.py
jerith: If it's been deleted or hasn't been stored yet. It's not a situation we expect, but we already handle it in the old code and there's no particular reason to change that.

vumi/components/message_store.py
jerith: `batch_ids` can be passed in by the caller.

jerith: Oh, wait. I see what you meant. Good catch. I'll add a test for this case.

vumi/components/message_store.py
```diff
@@ -1102,3 +1117,22 @@ def key_with_ts_and_value_formatter(batch_id, result):
...
+        # We already have batches, so there's no need to look them up.
+        returnValue(False)
+
+    outbound_messages = stored_event.manager.proxy(OutboundMessage)
+    msg_record = yield outbound_messages.load(stored_event.message.key)
+    if msg_record is not None:
+        for batch_id in msg_record.batches.keys():
+            stored_event.batches.add_key(batch_id)
+
+    returnValue(True)
```
justinvdm: We don't seem to use this function anywhere, how is it used?

jerith: Look at `vumi/scripts/vumi_model_migrator.py`. :-)

justinvdm: Ah, thanks :)

vumi/scripts/vumi_model_migrator.py
```diff
@@ -97,7 +108,8 @@ def migrate_key(self, key, dry_run):
         try:
             obj = yield self.model.load(key)
             if obj is not None:
-                if obj.was_migrated and not dry_run:
+                should_save = yield self.post_migrate_function(obj)
+                if should_save and not dry_run:
```
justinvdm: Why do we not need the `was_migrated` check anymore?

jerith: The default post-migrate function returns `obj.was_migrated`. If you provide your own post-migrate function, you can check `was_migrated` yourself if you care about it.

justinvdm: I noticed, but I would've thought that we only want to run the post-migrate function on objects that have been migrated. Seems a bit surprising that the post-migrate function would run on every object, regardless of whether or not it has been migrated.

vumi/components/tests/test_message_store.py
```diff
@@ -287,8 +292,65 @@ def test_add_ack_event(self):
...
+        timestamp = format_vumi_date(ack["timestamp"])
+        self.assertEqual(event.message_with_status, "%s$%s$ack" % (
+            msg_id, timestamp))
+        self.assertEqual(set(event.batches_with_statuses_reverse), set([
+            "%s$%s$ack" % (batch_1, to_reverse_timestamp(timestamp)),
+            "%s$%s$ack" % (batch_2, to_reverse_timestamp(timestamp)),
+        ]))
+
+    @inlineCallbacks
+    def test_add_ack_event_with_emtpy_batch_ids(self):
```
justinvdm: typo -- `emtpy` :)

vumi/scripts/vumi_model_migrator.py
jerith: Not necessarily. This is also useful for finding and fixing invalid data or any other case where we want to modify the content without necessarily changing the model version.

vumi/components/message_store.py
```diff
@@ -1102,3 +1119,22 @@ def key_with_ts_and_value_formatter(batch_id, result):
...
+
+@inlineCallbacks
+def add_batches_to_event(stored_event):
+    """
+    Post-migrate function to be used with `vumi_model_migrator` to add batches
+    to stored events that don't have any.
+    """
+    if stored_event.batches.keys():
+        # We already have batches, so there's no need to look them up.
+        returnValue(False)
```
justinvdm: Its a bit of a small edge case, but isn't it possible that we'd still want to re-add batches in case the migration crashed half-way through migrating a message?

vumi/scripts/vumi_model_migrator.py
justinvdm: Ah, cool.

justinvdm:
Left a minor comment, but :+1: once travis is happy. Lets maybe land on Monday though?

vumi/components/message_store.py
jerith: In that case, the stored event wouldn't have been saved and the next time it's loaded it'll still have no batches.

justinvdm: Ah, right.

PR #939, opened 2015-04-13
Split field and model tests better

jerith:
We currently have minimal field-specific tests in `vumi.persist.tests.test_fields` and a random assortment of model and field-specific tests in `vumi.persist.tests.test_model`. We should move all field-specific tests from the latter to the former.


jerith:
This is probably reviewed most easily one commit at a time.

vumi/persist/tests/test_fields.py
```diff
@@ -3,15 +3,117 @@
...
+
+
+class ModelFieldTestsDecorator(object):
+    """
+    Class decorator for replacing `@needs_riak`-marked test methods with two
+    wrapped versions, one for each Riak manager.
+
+    This is used here instead of the more usual mechanism of a mixin and two
+    subclasses because we have a lot of small test classes which have several
+    test methods that don't need Riak.
```
hodgestar: Couldn't we follow the usual pattern and just have decorator or call to `self.setup_riak()` in each test that needs it?

hodgestar: Hmm. I think I'll retract this since it's worth something to not have to remember to create subclasses of each test.

hodgestar:
I've mostly trusted that the tests were just moved across and not fiddled with too much. Generally looks good.

:+1:

vumi/persist/tests/test_fields.py
jerith: Yes. I started with the usual mixin-and-subclasses pattern and ended up drowning in boilerplate.

PR #941, opened 2015-05-21
Automatically release to PyPI if Travis builds of tags are successful.

hodgestar:


hodgestar:
Ready for review. Also, thoughts?

hodgestar:
Still keen to land this unless there are major objections?

rudigiesler:
No objections here, we probably need to update this too though: http://vumi.readthedocs.org/en/latest/how-we-do-releases.html

PR #942, opened 2015-05-25
Remove unnecessary message store indexes

jerith:
We currently have some unnecessary indexes on our message store objects, which may be contributing to Riak load and are definitely contributing to code complexity.


jerith:
This still needs model changes and all the accompanying migrations and tests.

hodgestar:
:+1: once Travis is happy.

PR #943, opened 2015-05-26
Allow model migrations to remove indexes

jerith:
Currently they can't.

### DIFF
--- a/vumi/components/message_store_migrators.py
+++ b/vumi/components/message_store_migrators.py
@@ -58,7 +58,7 @@ class EventMigrator(MessageMigratorBase):
         return mdata
 
     def reverse_from_2(self, mdata):
-        # We copy the `batches` field and related indexs even though the older
+        # We copy the `batches` field and related indexes even though the older
         # model version doesn't know about them. This lets us migrate
         # v2 -> v1 -> v2 without losing data.
         mdata.set_value('$VERSION', 1)
--- a/vumi/components/message_store_migrators.py
+++ b/vumi/components/message_store_migrators.py
@@ -53,17 +53,21 @@ class EventMigrator(MessageMigratorBase):
         mdata.set_value('batches', mdata.old_data.get('batches', []))
         mdata.copy_values('message')
         mdata.copy_indexes('message_bin')
+        mdata.copy_indexes('message_with_status_bin')
 
         return mdata
 
     def reverse_from_2(self, mdata):
-        # We copy the `batches` field even though the older model version
-        # doesn't know about it. This lets us migrate v2 -> v1 -> v2 without
-        # losing data.
+        # We copy the `batches` field and related indexs even though the older
+        # model version doesn't know about them. This lets us migrate
+        # v2 -> v1 -> v2 without losing data.
         mdata.set_value('$VERSION', 1)
         self._copy_msg_field('event', mdata)
         mdata.copy_values('message', 'batches')
         mdata.copy_indexes('message_bin')
+        mdata.copy_indexes('message_with_status_bin')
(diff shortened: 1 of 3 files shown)


jerith:
All model-related tests in vumi-go pass with this branch.

vumi/components/message_store_migrators.py
```diff
@@ -53,17 +53,21 @@ def migrate_from_1(self, mdata):
...
         mdata.copy_indexes('message_bin')
+        mdata.copy_indexes('message_with_status_bin')
 
         return mdata
 
     def reverse_from_2(self, mdata):
-        # We copy the `batches` field even though the older model version
-        # doesn't know about it. This lets us migrate v2 -> v1 -> v2 without
-        # losing data.
+        # We copy the `batches` field and related indexs even though the older
```
hodgestar: s/indexs/indexes/?

vumi/components/message_store_migrators.py
```diff
@@ -53,17 +53,21 @@ def migrate_from_1(self, mdata):
...
+        # model version doesn't know about them. This lets us migrate
+        # v2 -> v1 -> v2 without losing data.
```
hodgestar: I would actually expect the indexes to go away if we switched back to `v1` -- i.e. I'd expect data to be lost by the round trip. Not sure that's enough of a reason to actually through the indexes away. I assume that wanting current tests to pass is what prompted this change initially?

vumi/components/message_store_migrators.py
```diff
@@ -83,7 +87,7 @@ def migrate_from_1(self, mdata):
         mdata.set_value('$VERSION', 2)
         self._copy_msg_field('msg', mdata)
         mdata.copy_values('batches')
-        mdata.copy_indexes('batches')
+        mdata.copy_indexes('batches_bin')
```
hodgestar: This was a pre-existing bug that wasn't picked up because the old indexes were all copied anyway?

vumi/persist/tests/test_model.py
```diff
@@ -133,6 +133,43 @@ def reverse_from_3(self, migration_data):
...
+        # Migrator assertions
+        assert self.data_version == 3
+        assert self.model_class is IndexRemovedVersionedModel
+        assert isinstance(self.manager, Manager)
+
+        # Data assertions
+        assert set(migration_data.old_data.keys()) == set(
+            ['$VERSION', 'c', 'text'])
+        assert migration_data.old_data['$VERSION'] == 3
+        assert migration_data.old_index == {"text_bin": ["hi"]}
```
hodgestar: It seems odd to assert that ``text_bin`` is ``hi`` here because that means we can only pass in migration_data with ``text`` set to ``hi``. Maybe we should pass in what the old indexes are expected to be?

hodgestar: Or just leave this assertion out here and leave it up to the test?

hodgestar:
Left a few small comments, otherwise looks good.

vumi/components/message_store_migrators.py
jerith: Yes.

vumi/persist/tests/test_model.py
jerith: We can't make these assertions outside of the migration functions. Given that these migrations are written specifically for the tests and include any assertions at all, I don't think the more data-specific ones are entirely unreasonable.

vumi/components/message_store_migrators.py
jerith: The batch information can only come from outside the migration function, so throwing it away on a reverse migration is problematic.

hodgestar:
:+1:

PR #944, opened 2015-05-27
Computed value fields for models

jerith:
We currently do this very hackily in places and a cleaner solution would be good.


vumi/persist/fields.py
```diff
@@ -957,3 +963,27 @@ class ManyToMany(ForeignKey):
...
+    """A field descriptor for computed index fields."""
+
+    def pre_save(self, modelobj):
+        self.set_value(modelobj, self.field.value_func(modelobj))
+
+    def _add_index(self, modelobj, value):
+        # We override `FieldDescriptor._add_index` because we don't want `None`
+        # values.
+        if value is not None:
+            modelobj._riak_object.add_index(self.index_name, str(value))
```
hodgestar: How do indexes get removed?

jerith: `set_value()` clears the index values and then calls `_add_index()` as necessary.

vumi/persist/fields.py
```diff
@@ -957,3 +963,27 @@ class ManyToMany(ForeignKey):
...
+
+
+class ComputedIndex(Field):
+    """Field that stores a computed value for indexing."""
+
+    descriptor_class = ComputedIndexDescriptor
+
+    def __init__(self, value_func, index=True, **kw):
+        kw["null"] = True
+        super(ComputedIndex, self).__init__(index=index, **kw)
```
hodgestar: Could we perhaps just pass in `null=True` to `__init__` so that attempting to override `null` gives an error?

hodgestar: Does it make sense to allow `index=False`?

jerith: We always need `null=True` so we can create models without having values in the fields yet.

I'm not sure if `index=False` really makes sense. It doesn't in the current implementation, but I've been wondering if it's worth doing the computation whenever any field value changes. I've been leaning toward doing that, in which case this becomes a `ComputedValue` field that may or may not be indexed rather than a `ComputedIndex` field.

jerith:
This approach doesn't handle multiple index values, so I need to do something different here. Sorry. :-/

jerith:
This is ready for another review.

vumi/persist/fields.py
```diff
@@ -56,24 +57,50 @@ def _add_index(self, modelobj, value):
...
+    def get_riak_data(self, modelobj, default=None, key=None):
+        if key is None:
+            key = self.key
+        return modelobj._riak_object.get_data().get(key, default)
+
+    def set_riak_data(self, modelobj, raw_value, key=None):
+        if key is None:
+            key = self.key
+        old_raw_value = modelobj._riak_object.get_data().get(key)
+        modelobj._riak_object.set_data_field(key, raw_value)
```
JayH5: Why is it necessary to do a `set_data_field()` even if `old_raw_value == raw_value`?

jerith: Good point. I'll change that.

(This code originally lived at a higher level of abstraction, so we didn't actually know if the data would be modified by the write.)

jerith: Turns out we *do* actually need to set the value even if it's the same as we have before, because we want to store explicit `None` values instead of leaving the field out of the stored object.

vumi/persist/fields.py
```diff
@@ -56,24 +57,52 @@ def _add_index(self, modelobj, value):
...
+            key = self.key
+        return modelobj._riak_object.get_data().get(key, default)
+
+    def set_riak_data(self, modelobj, raw_value, key=None):
+        if key is None:
+            key = self.key
+        old_raw_value = modelobj._riak_object.get_data().get(key)
+        # We set this even if it's "unmodified" because we want explicit `None`
+        # values rather than missing fields.
+        modelobj._riak_object.set_data_field(key, raw_value)
```
JayH5: Makes me sad that we can't do any optimisation to save writes (even by special-casing `None`s?). Otoh, we don't know the performance benefit so I probably shouldn't worry so much...

jerith: These are in-memory operations on a dict, so they're about as cheap as it gets.

JayH5: Ah, so things are only written with `save()`? And then `save()` writes all the fields to Riak anyway? Or are there checks there to see if values have changed?

jerith: Yes. The whole object is a single entity in Riak and it gets written unconditionally whenever `save()` is called.

Trying to detect changes (which could be to indexes or metadata rather than just the object itself) is more complexity than we want, especially since writes of unmodified objects are much less common than writes of modified objects. The only reason we're bothering with it here is to avoid loops where computed values trigger each other's update functions despite not actually depending on each other.

JayH5:
:+1: from me :)

hodgestar:
Why don't we just update the value on `get` or `save` instead of every time someone updates any field on the model?

jerith:
> Why don't we just update the value on `get` or `save` instead of every time someone updates any field on the model?

I tried that before switching to update-on-value-change and ran into problems trying to intercept all accesses to object data (including things that look into the the underlying Riak object directly) to make sure updates didn't get missed. It ended up being cleaner doing it this way.

hodgestar:
>  I tried that before switching to update-on-value-change and ran into problems trying to intercept all   accesses to object data (including things that look into the the underlying Riak object directly) to make sure updates didn't get missed. It ended up being cleaner doing it this way.

But the only `get` we need to worry about is getting the computed value itself? Having a pre-save hook might also allow us to add cross-field validation later, which would be cool.

jerith:
> But the only `get` we need to worry about is getting the computed value itself? Having a pre-save hook might also allow us to add cross-field validation later, which would be cool.

We need to make sure that everything we do with the object data, including the low-level operations that bypass fields (saving, migration, etc.) has the right values and metadata for all fields. Hooking into the various things that read that is harder than triggering updates on writes through the model object. (As I said, I tried this and it ended up being more complicated than I was happy with.)

More importantly, it's a good idea to catch validation or computation errors when the data is modified instead of some future time when we try to read a field or save the object.

hodgestar:
One can't do cross-field validation when a single field is updated.

jerith:
This PR isn't about cross-field validation. Even if it were, we'd want the validation to happen before the update and not at the next read or save at some arbitrary point in the future.

hodgestar:
:+1:

PR #945, opened 2015-06-02
Structured logging

jerith:
Our log messages are currently very ad-hoc and really hard to post-process in any useful way.

Using [eliot](http://eliot.readthedocs.org) we can log structured data that's a little harder for humans to read but much easier to reassemble into useful information about message flows and such.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ setup(
         'treq',
         'confmodel>=0.2.0',
         'hyperloglog',
+        'eliot>=0.7.1',
     ],
     classifiers=[
         'Development Status :: 4 - Beta',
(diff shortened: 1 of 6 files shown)


jerith:
This is a very preliminary attempt to use eliot for some logging that would be immediately useful to us.

We still need to figure out what we want to do with the log messages (they currently go nowhere except in tests) and how we want to extend this logging to other parts of the system.

jerith:
Oh, bleh. Eliot doesn't support Python 2.6.

hodgestar:
Do we need to still support Python 2.6? It's pining for fjords. :)

jerith:
I've removed it from this branch, at least. We can decide later.

PR #946, opened 2015-06-03
Remove Riak empty index hack

jerith:
We are still storing `"None"` for empty index values, but the need for this has long since passed.

### DIFF
--- a/vumi/persist/fields.py
+++ b/vumi/persist/fields.py
@@ -9,6 +9,16 @@ from vumi.message import format_vumi_date, parse_vumi_date
 from vumi.utils import to_kwargs
 
 
+# Index values in Riak have to be non-empty, so a zero-length string
+# counts as "no value". Since we still have legacy data that was
+# inadvertantly indexed with "None" because of the str() call in the
+# library and we still have legacy code that relies on an index search
+# for a value of "None", fixing this properly here will break existing
+# functionality. Once we have rewritten the offending code to not use
+# "None" in the index, we can remove the hack below and be happier.
+STORE_NONE_FOR_EMPTY_INDEX = False
+
+
 class ValidationError(Exception):
     """Raised when a value assigned to a field is invalid."""
 
@@ -39,18 +49,11 @@ class FieldDescriptor(object):
     def _add_index(self, modelobj, value):
         # XXX: The underlying libraries call str() on whatever index values we
         # provide, so we do this explicitly here and special-case None.
-        #
-        # Index values in Riak have to be non-empty, so a zero-length string
-        # counts as "no value". Since we still have legacy data that was
-        # inadvertantly indexed with "None" because of the str() call in the
-        # library and we still have legacy code that relies on an index search
-        # for a value of "None", fixing this properly here will break existing
-        # functionality. Once we have rewritten the offending code to not use
-        # "None" in the index, we can remove the hack below and be happier.
         if value is None:
             value = ''
-            # FIXME: We still rely on this being "None" in places. :-(
-            value = 'None'
+            if STORE_NONE_FOR_EMPTY_INDEX:
+                # FIXME: We still rely on this being "None" in places. :-(
+                value = 'None'
         modelobj._riak_object.add_index(self.index_name, str(value))
 
     def set_value(self, modelobj, value):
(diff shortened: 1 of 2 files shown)


jerith:
All vumi-go tests pass locally for me with this change.

vumi/persist/fields.py
```diff
@@ -39,18 +49,11 @@ def initialize(self, modelobj, value):
...
-        # functionality. Once we have rewritten the offending code to not use
-        # "None" in the index, we can remove the hack below and be happier.
         if value is None:
             value = ''
-            # FIXME: We still rely on this being "None" in places. :-(
-            value = 'None'
+            if STORE_NONE_FOR_EMPTY_INDEX:
+                # FIXME: We still rely on this being "None" in places. :-(
+                value = 'None'
         modelobj._riak_object.add_index(self.index_name, str(value))
```
hodgestar: I was thinking that we could not add an index here at all in the `None` case. I.e. treat None the same as not having an indexable value?

vumi/persist/tests/test_model.py
```diff
@@ -560,6 +561,25 @@ def test_index_keys(self):
...
+        self.patch(fields, "STORE_NONE_FOR_EMPTY_INDEX", True)
+        indexed_model = self.manager.proxy(IndexedModel)
+        yield indexed_model("foo1", a=1, b=u"one").save()
+        yield indexed_model("foo2", a=2, b=u"one").save()
+        yield indexed_model("foo3", a=2, b=None).save()
+
+        keys = yield indexed_model.index_keys('a', 1)
+        self.assertEqual(keys, ["foo1"])
+        # We should get a list object, not an IndexPage wrapper.
+        self.assertTrue(isinstance(keys, list))
```
hodgestar: Are the isinstance and `.index_keys('a', 1)` relevant to this test? We really just want to check that looking up `b=None` and `b="NotNone"` are okay?

vumi/persist/tests/test_model.py
```diff
@@ -576,6 +596,23 @@ def test_index_keys_return_terms(self):
...
+        yield indexed_model("foo2", a=2, b=u"one").save()
+        yield indexed_model("foo3", a=2, b=None).save()
+
+        keys = yield indexed_model.index_keys('a', 1, return_terms=True)
+        self.assertEqual(keys, [("1", "foo1")])
+
+        keys = yield indexed_model.index_keys('b', u"one", return_terms=True)
+        self.assertEqual(sorted(keys), [(u"one", "foo1"), (u"one", "foo2")])
+
+        keys = yield indexed_model.index_keys('b', None)
```
hodgestar: Do we want `return_terms=True` here?

vumi/persist/tests/test_model.py
```diff
@@ -576,6 +596,23 @@ def test_index_keys_return_terms(self):
...
+    @Manager.calls_manager
+    def test_index_keys_return_terms_store_none_for_empty(self):
+        self.patch(fields, "STORE_NONE_FOR_EMPTY_INDEX", True)
+        indexed_model = self.manager.proxy(IndexedModel)
+        yield indexed_model("foo1", a=1, b=u"one").save()
+        yield indexed_model("foo2", a=2, b=u"one").save()
+        yield indexed_model("foo3", a=2, b=None).save()
+
+        keys = yield indexed_model.index_keys('a', 1, return_terms=True)
+        self.assertEqual(keys, [("1", "foo1")])
```
hodgestar: Also wondering whether this check is relevant to this test?

hodgestar:
Left a few more comments. Otherwise looks good.

vumi/persist/tests/test_model.py
jerith: These tests are unmodified copies of what we had before except for the monkey patch at the top. I think it's still worth testing all of the original cases while we have the hack in the codebase, because the tests shouldn't assume that setting `STORE_NONE_FOR_EMPTY_INDEX = True` won't affect non-`None` values.

vumi/persist/tests/test_model.py
jerith: We do. That was a bug in the original test. Good catch. :-)

hodgestar:
:+1:

PR #947, opened 2015-06-03
When storing events, fallback to an existing events idea of the batches first.

hodgestar:
Currently we always look up the outbound message.

### DIFF
--- a/vumi/components/message_store.py
+++ b/vumi/components/message_store.py
@@ -424,26 +424,33 @@ class MessageStore(object):
         returnValue(msg.msg if msg is not None else None)
 
     @Manager.calls_manager
+    def _get_batches_from_outbound(self, msg_id):
+        msg_record = yield self.outbound_messages.load(msg_id)
+        if msg_record is not None:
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

vumi/components/message_store.py
```diff
@@ -424,26 +424,33 @@ def get_outbound_message(self, msg_id):
...
         event_id = event['event_id']
         msg_id = event['user_message_id']
         event_record = yield self.events.load(event_id)
         if event_record is None:
             event_record = self.events(event_id, event=event, message=msg_id)
+            if batch_ids is None:
+                # If we aren't given batch_ids, get them from the outbound
+                # message.
+                batch_ids = yield self._get_batches_from_outbound(msg_id)
+                print batch_ids
```
smn: stray print

smn:
:+1: from me but I'm a bit behind on these internals.

jerith:
:+1: based on local test runs. Travis builds are stalled and it'll be a while before we see any happen. :-/

PR #948, opened 2015-06-04
TxRiakManager needs its own threadpool

jerith:
It currently uses the general reactor threadpool, but we want tighter control over concurrency and we don't want to muck about with the reactor's threadpool in case other things are using it.

### DIFF
--- a/vumi/persist/txriak_manager.py
+++ b/vumi/persist/txriak_manager.py
@@ -5,7 +5,8 @@
 import json
 
 from riak import RiakClient, RiakObject, RiakMapReduce, RiakError
-from twisted.internet.threads import deferToThread
+from twisted.python.threadpool import ThreadPool
+from twisted.internet.threads import deferToThreadPool
 from twisted.internet.defer import (
     inlineCallbacks, returnValue, gatherResults, maybeDeferred, succeed)
 
@@ -34,8 +35,9 @@ class VumiTxIndexPage(object):
     Iterating over this object will return the results for the current page.
     """
 
-    def __init__(self, index_page):
+    def __init__(self, index_page, manager):
         self._index_page = index_page
+        self._manager = manager
 
     def __iter__(self):
         if self._index_page.stream:
@@ -71,15 +73,16 @@ class VumiTxIndexPage(object):
         """
         if not self.has_next_page():
             return succeed(None)
-        d = deferToThread(self._index_page.next_page)
-        d.addCallback(type(self))
+        d = self._manager.deferToThread(self._index_page.next_page)
+        d.addCallback(type(self), self._manager)
         d.addErrback(riakErrorHandler)
         return d
 
 
 class VumiTxRiakBucket(object):
-    def __init__(self, riak_bucket):
+    def __init__(self, riak_bucket, manager):
         self._riak_bucket = riak_bucket
+        self._manager = manager
 
     def get_name(self):
         return self._riak_bucket.name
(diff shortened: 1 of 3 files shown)


vumi/persist/txriak_manager.py
```diff
@@ -180,6 +184,16 @@ class TxRiakManager(Manager):
 
     call_decorator = staticmethod(inlineCallbacks)
 
+    def _setup_manager(self):
+        from twisted.internet import reactor
+        self._reactor = reactor
+        self._threadpool = ThreadPool(minthreads=2, maxthreads=2)
```
hodgestar: Wondering what a good number of threads is.

vumi/persist/txriak_manager.py
```diff
@@ -180,6 +184,16 @@ class TxRiakManager(Manager):
...
+        self._threadpool.start()
```
hodgestar: Do we need to stop the threadpool at some point?

hodgestar:
And thoughts on how we can test that the thread pool is being used where its supposed to be?

vumi/persist/txriak_manager.py
jerith: Right, good point. I forgot to put that in.

vumi/persist/txriak_manager.py
rudigiesler: Is 2 threads a good number? Is there a way we can set this in a config somewhere so that we can adjust it to see its effect?

jerith: 2 threads is a guess, but I don't think it's an unreasonable one. Making it configurable is tricky because it means having different configuration options between the different Riak managers. It's definitely something we should do, but I'd like to get this out first to see if it solves some of our immediate problems.

jerith:
Builds failed because the threads weren't being killed and they're not daemon threads, so they block process exit.

jerith:
Ready for another review.

vumi/persist/txriak_manager.py
```diff
@@ -155,31 +159,60 @@ def set_user_metadata(self, usermeta):
...
-        d.addCallback(type(self))
+        d = self._manager.deferToThread(self._riak_obj.delete)
+        d.addCallback(type(self), self._manager)
         return d
 
 
+def _daemon_threadpool(*args, **kw):
+    """
+    Create a threadpool that sets its threads to daemon mode so they don't
+    block process exit if they aren't properly cleaned up.
```
hodgestar: Don't we want process exit to wait until riak requests are done?

jerith: The more relevant question here is whether we want to hang forever because something didn't properly close a `TxRiakManager` object.

Calling `TxRiakManager.close_manager()` will wait for the threads to finish what they're doing and stop them. If that doesn't happen, idle threads will hang around forever and if they're non-daemon threads they'll keep the process alive past the termination of the main thread.

hodgestar: Wouldn't SIGKILL still kill everything? It's nice to have SIGTERM end a
process cleanly.

jerith: This has nothing to do with whether there's an incomplete Riak operation during shutdown, it's about what happens to threads that are not explicitly cleaned up.

If you fail to clean up a manager with non-daemon threads, you're guaranteed to have the process hang indefinitely (unless you hit it with SIGKILL) because the still-running threadpool will have threads that are waiting for more work to do. An in-progress Riak operation could still complete after the main thread terminates, but it would have nowhere to send its result and its thread will still hang around afterwards waiting for more work that will never arrive.

If you fail to clean up a manager with daemon threads, there's a risk of killing an in-progress Riak operation that you're not waiting for during service shutdown. This is exactly the same as a completely async operation where you're not waiting for things to finish during service shutdown.

jerith: I should note that I discovered this stuff because my test run processes never terminated and I had to `kill -9` them. Consider that behaviour in production when a worker process fails or is restarted.

hodgestar: Supervisor does a `kill -9` eventually anyway -- one sort of has to when managing arbitrary processes.

I'm worried that we'll lose results and not know because the process appeared to exist cleanly. Not disagreeing that things are broken in both cases.

hodgestar:
:+1:

vumi/persist/txriak_manager.py
jerith: The place to handle this is our normal shutdown process management, which we already do.

PR #950, opened 2015-06-05
Ensure Riak managers are properly cleaned up

jerith:
At the moment we don't (and we also don't properly clean them up) which is potentially a very large problem.


jerith:
These assertions are currently optional because vumi-go still leaves unclosed Riak managers all over the place. We need to do some more work to make that cleanup possible.

vumi/middleware/tests/test_message_storing.py
```diff
@@ -15,7 +15,8 @@ def setUp(self):
 
         # Create and stash a riak manager to clean up afterwards, because we
         # don't get access to the one inside the middleware.
-        self.persistence_helper.get_riak_manager()
+        manager = self.persistence_helper.get_riak_manager()
```
hodgestar: Should the persistence_helper not close the manager for us?

hodgestar:
Should we be doing the same checks for unclosed redis managers?

hodgestar:
Left two comments, otherwise looks good.

vumi/middleware/tests/test_message_storing.py
jerith: There are places where we patch in `persistence_helper.get_riak_manager()` to replace the code under test's own manager creation, and we need to make sure that those also do their own cleanup.

jerith:
We should do the same for redis, but that's out of scope for this PR.

vumi/middleware/tests/test_message_storing.py
hodgestar: But we're immediately calling ``add_cleanup(manager.close)`` here so we're not checking anyway? Not entirely sure I understood your response to my comment.

smn:
:+1: from me but :+1: on @hodgestar doing another review.

vumi/middleware/tests/test_message_storing.py
jerith: I'm talking about the tests where we actually replace the code-under-test's Riak manager setup with our own -- `test_vumi_model_migrator.py`, for example.

Except in those places we're just handing out copies of our own manager instead of creating new ones, which means we're still not testing that those things properly close their connections. I'll fix that.

hodgestar: Tx.

jerith:
Ready for another review.

jerith:
The failed Travis build is a weird pypy issue that I'd like to look into further, so please don't restart it. All other build jobs were fine.

jerith:
Never mind. That's apparently a bug which has been fixed since pypy 2.5.0 (which is what Travis has).

vumi/tests/helpers.py
```diff
@@ -1268,6 +1276,13 @@ def cleanup(self):
 
         self._unpatch()
 
+        if unclosed_managers and self._assert_closed:
+            # We have unclosed managers and we've been asked to assert that we
+            # don't.
+            raise Exception(
```
hodgestar: This should probably have a custom exception class, I guess.

hodgestar:
I'd still like us to clean up the creation of riak managers in tests. Our persistence test helpers can already be very complicated to use and having to manually add a cleanup call to `manager.close_manager` is making that burden worse.

I think the underlying cause of the trouble might be that there are two distinct use cases:

1. Retrieving a riak manager for direct use in a test, or to pass to a function or class that takes a riak manager and isn't expected to do its own clean up.  In this case we want the persistence helper to clean up the manager for us.

2. Creating a riak manager on behalf of a worker or similar that is expected to clean up its own manager.

Perhaps it would be easiest to support these if the persistence helper had two different methods for retrieving the riak manager, or took a keyword argument that specifies whether to clean-up the manager automatically?

jerith:
The persistence helper doesn't have an easy way to determine which things need cleanup and which don't, so adding the cleanup will require some effort. Also, I don't want to make cleaning up the manager the default behaviour, because that makes it too easy to unintentionally create a self-closing manager and then not actually test that it's being closed properly. It's better to accidentally write tests that fail when they shouldn't than it is to write tests that pass when they shouldn't.

In any case, I don't believe requiring manual cleanup of explicitly created manager in tests is much of a burden. If you write a test that creates a manager without adding the cleanup and then run it, you'll get a test failure indicating that you have an unclosed manager. Having just written that code that creates said manager, the cause should be pretty clear.

vumi/tests/helpers.py
jerith: Goog point. That'll make the test for this behaviour more specific, if nothing else.

jerith:
While trying to get vumi-go tests to pass with `VUMI_TEST_ASSERT_CLOSED=true`, I had a hard time finding all the places unclosed managers were created. (Some tests had tens or hundreds.)

I added some code to store a stack trace during manager creation and display it for any unclosed manager, which makes it much easier to find the managers we're not closing properly.

jerith:
That said, vumi-go makes heavy use of temporary sub-managers. Since sub-managers already share their parent's client object, cleanup (and threadpool management, etc.) should happen at that level rather  than per-manager. I'm going to take a stab at that and see if that leads to a cleaner solution.

hodgestar:
Sounds good.

jerith:
Closing in favour of #957.

PR #953, opened 2015-06-10
Configurable reactor threadpool size

jerith:
Giving each Riak manager its own threadpool is complicated for a variety of reasons, so being able to configure the reactor's threadpool size is useful.

### DIFF
--- a/vumi/servicemaker.py
+++ b/vumi/servicemaker.py
@@ -119,6 +119,7 @@ class StartWorkerOptions(VumiOptions):
         ["worker_class", None, None, "Deprecated. See --worker-class instead"],
         ["config", None, None, "YAML config file for worker configuration"
          " options"],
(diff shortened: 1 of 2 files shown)


vumi/servicemaker.py
```diff
@@ -197,6 +203,12 @@ class VumiWorkerServiceMaker(object):
     # what command line options does this service expose
     options = StartWorkerOptions
 
+    def set_maxthreads(self, options):
```
smn: Why are we passing in the options object here instead of the `maxthreads` value?

smn:
Other than the bikeshed question about the options argument, :+1: once travis.

vumi/servicemaker.py
jerith: That method was originally doing something more complicated. I'll make the change, thanks.

hodgestar:
:+1: once Travis is happy.

PR #954, opened 2015-06-15
Better Riak management in script tests

jerith:
Trying to properly clean up Riak managers (and assert that the cleanup is happening) has uncovered problems with how we handle Riak managers in various scripts. This is separate from the work in #950, but necessary for it.

### DIFF
--- a/vumi/tests/helpers.py
+++ b/vumi/tests/helpers.py
@@ -1399,7 +1399,7 @@ class PersistenceHelper(object):
                 "setup() must be called before performing this operation.")
 
     @proxyable
-    def get_riak_manager(self, config=None):
+    def get_riak_manager(self, config=None, prepend_bucket_prefix=True):
         """
         Build and return a Riak manager.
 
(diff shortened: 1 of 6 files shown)


vumi/tests/helpers.py
```diff
@@ -1407,14 +1407,25 @@ def get_riak_manager(self, config=None):
...
             value of :attr:`is_sync`.
         """
         self._check_patches_applied()
+        default_config = self._config_overrides["riak_manager"].copy()
         if config is None:
-            config = self._config_overrides['riak_manager'].copy()
+            config = default_config
+        else:
+            config = config.copy()
+            if prepend_bucket_prefix:
```
hodgestar: What is the use case for not appending the default bucket prefix?

hodgestar:
Left one question, otherwise looks good.

vumi/tests/helpers.py
jerith: I haven't found one, but I didn't look at the vumi-go code at all. The option is mostly there to let us get the original behaviour back if necessary.

hodgestar: Could we just leave the option out for now and add it back if we really
need it? I think any tests currently relying on the old behaviour should be
fixed since having a bucket prefix outside of the test one seems like it
might cause problems if, e.g., riak on a dev's machine has other data in it.

jerith: Agreed, I'll do that.

jerith:
Ready for another review.

hodgestar:
:+1:

PR #955, opened 2015-06-18
Fix riak mapreduce timeout test

jerith:


rudigiesler:
:+1: when travis is happy

vumi/persist/tests/test_txriak_manager.py
```diff
@@ -217,19 +206,22 @@ def test_run_riak_map_reduce_with_timeout(self):
...
         try:
             yield self.manager.run_map_reduce(mr, lambda m, l: None)
         except Exception, err:
-            msg = unrepr_string(str(err))
-            self.assertTrue(msg.startswith(
-                "Error running MapReduce operation."))
-            self.assertTrue(msg.endswith(
-                "Body: '{\"error\":\"timeout\"}'"))
+            msg = str(err)[1:-1].decode("string-escape")
+            if not all([
```
JayH5: Why not just an and?

rudigiesler:
:+1:

vumi/persist/tests/test_txriak_manager.py
jerith: I find the `all()` style a little cleaner than `(foo and bar)` when the expression spans multiple lines. The short-circuiting of the `and` operator wouldn't buy much here, so it basically just boils down to a mild readability preference. Both forms work and I'm happy to change it if you think the `and` works better -- I'm never really sure whether this kind of readability preference is a personal quirk of mine or something more general.

JayH5: I actually quite like the `all()` style because you can think of it as a list of conditions to be satisfied. Otoh, `and` is simpler. Don't want to get stuck on this tiny detail - I was mostly just curious if this was a common Python thing people did...

JayH5:
:+1:

PR #956, opened 2015-06-24
Factor out common Riak wrapper code

jerith:
`RiakManager` and `TxRiakManager` both have a lot of wrapper code around the underlying library objects, and there's a bunch of duplication here. This issue is for factoring out the common bits and adding a new wrapper around the client object that we can add extra things to later.

(This is another case of pulling a piece of the big Riak manager cleanup/threadpool/whatever mudball out into its own smaller change.)

### DIFF
--- a/vumi/persist/riak_base.py
+++ b/vumi/persist/riak_base.py
@@ -5,13 +5,13 @@ import json
 from riak import RiakClient
 
 
-def to_unicode(text, encoding='utf-8'):
+def _to_unicode(text, encoding='utf-8'):
     # If we already have unicode or `None`, there's nothing to do.
     if isinstance(text, (unicode, type(None))):
         return text
     # If we have a tuple, we need to do our thing with every element in it.
     if isinstance(text, tuple):
-        return tuple(to_unicode(item, encoding) for item in text)
+        return tuple(_to_unicode(item, encoding) for item in text)
     # If we get here, then we should have a bytestring.
     return text.decode(encoding)
 
@@ -86,7 +86,7 @@ class VumiIndexPageBase(object):
     def __iter__(self):
         if self._index_page.stream:
             raise NotImplementedError("Streaming is not currently supported.")
-        return (to_unicode(item) for item in self._index_page)
+        return (_to_unicode(item) for item in self._index_page)
 
     def __eq__(self, other):
         return self._index_page.__eq__(other)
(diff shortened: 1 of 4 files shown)


jerith:
This is ready for review. Any suggestions for other things to factor out would also be helpful.

hodgestar:
Looks sane to me. Do we want tests for things in the base classes?

jerith:
We already have common tests that run across both subclasses, so I don't think tests on the base classes specifically would add much.

vumi/persist/riak_manager.py
```diff
@@ -97,69 +67,22 @@ def get_index_page(self, index_name, start_value, end_value=None,
...
-        return type(self)(self._riak_obj.reload())
-
-    def delete(self):
-        return type(self)(self._riak_obj.delete())
+    def _call_and_wrap(self, func):
+        """
+        Call a function that touches the network and wrap the result in this
+        class.
+        """
+        return type(self)(func())
```
JayH5: My python knowledge is being stretched here... but why can't this (`return type(self)(func())`) be in the base class?

vumi/persist/txriak_manager.py
```diff
@@ -104,73 +74,22 @@ def get_index_page(self, index_name, start_value, end_value=None,
...
-        return d
-
-    def delete(self):
-        d = deferToThread(self._riak_obj.delete)
+    def _call_and_wrap(self, func):
+        """
+        Call a function that touches the network and wrap the result in this
+        class.
+        """
+        d = deferToThread(func)
```
JayH5: Ah, never mind... now I see

vumi/persist/riak_base.py
```diff
@@ -0,0 +1,208 @@
...
+
+from riak import RiakClient
+
+
+def to_unicode(text, encoding='utf-8'):
+    if text is None:
+        return text
+    if isinstance(text, tuple):
+        return tuple(to_unicode(item, encoding) for item in text)
+    if not isinstance(text, unicode):
```
JayH5: Super minor nitpick... but I find this logic a bit confusing. Why not...
```python
if text is None or isinstance(text, unicode):
    return text
...
return text.decode(encoding)
```
... just the `is not` and then the sort of "fall-through" of the unicode case I found unclear.

jerith: The short and unsatisfactory answer is "I just moved the existing code without modifying it". The longer answer is that the current implementation probably had the tuple case added later and the structure made more sense before that.

I'll take a look at this and see if there's a better way to write this function, thanks.

JayH5:
:+1: :ship:

PR #957, opened 2015-06-25
Ensure Riak clients properly cleaned up (#950)

jerith:
This is a reimplementation of #950.

### DIFF
--- a/vumi/persist/riak_manager.py
+++ b/vumi/persist/riak_manager.py
@@ -122,7 +122,17 @@ class RiakManager(Manager):
             mapreduce_timeout=mapreduce_timeout, store_versions=store_versions)
 
     def close_manager(self):
-        self.client.close()
+        if self._parent is None:
+            # Only top-level managers may close the client.
+            self.client.close()
+
+    def _is_unclosed(self):
+        # This returns `True` if the manager needs to be explicitly closed and
+        # hasn't been closed yet. It should only be used in tests that ensure
+        # client objects aren't leaked.
+        if self._parent is not None:
+            return False
+        return not self.client._closed
 
     def riak_bucket(self, bucket_name):
         bucket = self.client.bucket(bucket_name)
--- a/vumi/scripts/vumi_model_migrator.py
+++ b/vumi/scripts/vumi_model_migrator.py
@@ -87,8 +87,8 @@ class ModelMigrator(object):
         riak_config = {
             'bucket_prefix': options['bucket-prefix'],
         }
-        manager = self.get_riak_manager(riak_config)
-        self.model = manager.proxy(model_cls)
+        self.manager = self.get_riak_manager(riak_config)
+        self.model = self.manager.proxy(model_cls)
 
         # The default post-migrate-function does nothing and returns True if
         # and only if the object was migrated.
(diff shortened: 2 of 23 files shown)


vumi/persist/riak_manager.py
```diff
@@ -122,7 +122,17 @@ def from_config(cls, config):
             mapreduce_timeout=mapreduce_timeout, store_versions=store_versions)
 
     def close_manager(self):
-        self.client.close()
+        if self._parent is None:
+            # Only top-level managers may close the client.
+            self.client.close()
```
rudigiesler: Should we raise an exception if we don't close the client because it's not top level? Seems a bit strange to call `close_manager` and for it not to close anything, or maybe that's what should happen. Just a thought.

jerith: Logically, `close_manager()` means "I'm finished with this manager" rather than "clean up all the underlying resources". The thing that calls it doesn't necessarily know (or care) whether it has a top-level manager or not.

My initial implementation required sub-managers to be explicitly closed as well, but there are too many places (mostly in vumi-go) where we cache model objects or otherwise have sub-managers with indeterminate lifecycles.

rudigiesler:
I'm liking the idea of keeping the `_parent` on child objects. Makes the implementation quite clean.

vumi/tests/helpers.py
```diff
@@ -169,6 +171,43 @@ def get_timeout():
...
+
+    filtered_stack = list(dropwhile(is_boring, stack))
+    if filtered_stack:
+        # We haven't accidentally devoured everything.
+        lines_removed = len(stack) - len(filtered_stack)
+        stack = filtered_stack
+        stack.insert(0, "%s test runner lines removed.\n" % lines_removed)
+    # Remove the current stack frame and any extra stack frames we've been
+    # asked to remove.
+    return stack[:-(exclude_last + 1)]
```
rudigiesler: Wouldn't this mean that the `%s test runner lines removed` line might be removed?

rudigiesler: Nevermind, I didn't see that it's an insert. Maybe we want to have something similar at the end, to show that lines have been removed? (If they have been removed)

jerith: The lines we're removing from the end are always going to be noise from the calls we're making to get the stack trace. The lines we're removing from the beginning are our best guess at what the test runner spam might be, so it's helpful to at least note that we're removing them in case we guess wrong or something.

rudigiesler:
:+1: from me. I'm liking this implementation, it seems much cleaner.

vumi/scripts/vumi_model_migrator.py
```diff
@@ -186,12 +189,15 @@ def migrate_all_keys(self, continuation=None):
...
 
-    def run(self):
+    def _run(self):
         if self.options["keys"] is not None:
             return self.migrate_specified_keys(self.options["keys"].split(","))
         else:
             return self.migrate_all_keys(self.options["continuation-token"])
 
+    def run(self):
+        return self._run().addBoth(lambda _: self.cleanup())
```
hodgestar: Is there a reason this `run` implementation uses `addBoth` and the other one uses `inlineCallbacks`?

jerith: Probably because I wrote them at different times. I'll change them to be consistent -- do you have any preference?

hodgestar:
Looks okay to me too. :+1:

vumi/scripts/vumi_model_migrator.py
hodgestar: No preference -- just wanted to check that I wasn't missing a reason. Also
happy for them to stay as they are.

PR #958, opened 2015-07-08
Implement `__len__()` in wrapped Riak page objects

JayH5:
The underlying Riak client library objects implement `__len__()` so this should be possible. This should be a small optimisation for counting the length of items in a page (which doesn't happen that often).

Class of interest is `vumi.persist.riak_base.VumiIndexPageBase`


jerith:
:+1:

PR #960, opened 2015-08-20
Accept empty SMSes and pass them on as empty strings.

hodgestar:
It appears that some phones and networks support sending empty SMSes. We currently reject those but should probably forward them on as empty messages.


hodgestar:
Ready for review.

jerith:
:+1:
I'll fix the problem with cryptography and old pypy in a separate PR.

PR #961, opened 2015-08-21
Disallow empty SMSes by default.

hodgestar:
After some discussion with others (@smn, @jerith) it seems like a good idea to default to disallowing empty SMSes.


hodgestar:
Ready for review.

rudigiesler:
:+1: , travis failure seems to be the same pypy issue as before

PR #962, opened 2015-08-21
Old cryptography for old pypy

jerith:
`cryptography>=1.0` only supports pypy 2.6.0 and later. Travis has an older pypy than that, so builds fail.

The plan is to detect old pypy in `setup.py` and add a constraint on the `cryptography` version.


hodgestar:
:+1:

PR #963, opened 2015-08-25
Fix race condition in dethrottling SMPP messages

hodgestar:
The ordering of these two lines at the end of `handle_outbound message` potentially allow a message to be throttled and unthrottled before it has been stored in the message cache::
```
        yield self.submit_sm_processor.handle_outbound_message(
            message, protocol)
        yield self.message_stash.cache_message(message)
```
at https://github.com/praekelt/vumi/blob/develop/vumi/transports/smpp/smpp_transport.py#L407-L409

### DIFF
--- a/vumi/transports/smpp/smpp_transport.py
+++ b/vumi/transports/smpp/smpp_transport.py
@@ -404,9 +404,9 @@ class SmppTransceiverTransport(Transport):
         if not self._check_address_valid(message, 'from_addr'):
             yield self._reject_for_invalid_address(message, 'from_addr')
             return
+        yield self.message_stash.cache_message(message)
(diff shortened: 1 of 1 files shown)


hodgestar:
I don't have a test for this but I'm fairly sure this ordering is better.

Writing a test would be a bit tricky -- we'd need to send a first message and throttle it but pause just before storing it in the cache. Then we'd need to send a second message and have that trigger an attempt to unthrottle the first message. Happy for someone else to give this a go if they're keen.

hodgestar:
Ready for review (or at least comments).

jerith:
I spent some time this morning trying (unsuccessfully) to come up with a test for this. The change itself is innocuous enough, so I don't see any harm in making it.

:+1:

PR #967, opened 2015-09-04
Fix Riak race condition in test

jerith:
`TestMessageStoreResource.test_disconnect_kills_server` leaves some unfinished Riak operations running, which results in a race condition on cleanup and occasional build failures.


smn:
:+1:

PR #968, opened 2015-09-07
Better SMPP multipart_info handling

jerith:
At the moment, this uses the week-long `third_party_id_expiry` and is never cleared. Instead it should use the day-log `submit_sm_expiry` and be cleared after all parts of a message have been acknowledged.

(There may be some overlap with #965, but running out of memory in Redis is a worse problem than potential extra acks.)

### DIFF
--- a/vumi/transports/smpp/config.py
+++ b/vumi/transports/smpp/config.py
@@ -37,6 +37,11 @@ class SmppTransportConfig(Transport.CONFIG_CLASS):
         'for matching submit_sm_resp and delivery report messages. Defaults '
         'to 1 week.',
         default=(60 * 60 * 24 * 7), static=True)
+    completed_multipart_info_expiry = ConfigInt(
+        'How long (in seconds) to keep multipart message info for completed '
+        'multipart messages around to avoid pending operations accidentally '
+        'recreating them without an expiry time. Defaults to 1 hour.',
+        default=(60 * 60), static=True)
     redis_manager = ConfigDict(
         'How to connect to Redis.', default={}, static=True)
     split_bind_prefix = ConfigText(
--- a/vumi/transports/smpp/smpp_transport.py
+++ b/vumi/transports/smpp/smpp_transport.py
@@ -421,10 +421,10 @@ class SmppTransceiverTransport(Transport):
     def process_submit_sm_event(self, message_id, event_type, remote_id,
                                 command_status):
         if event_type == 'ack':
-            if not self.disable_ack:
-                yield self.publish_ack(message_id, remote_id)
             yield self.message_stash.delete_cached_message(message_id)
             yield self.message_stash.expire_multipart_info(message_id)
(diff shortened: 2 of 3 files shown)


vumi/transports/smpp/config.py
```diff
@@ -37,6 +37,11 @@ class SmppTransportConfig(Transport.CONFIG_CLASS):
         'for matching submit_sm_resp and delivery report messages. Defaults '
         'to 1 week.',
         default=(60 * 60 * 24 * 7), static=True)
+    completed_multipart_info_expiry = ConfigInt(
+        'How long (in seconds) to keep multipart message info for completed '
+        'multipart messages around to avoid pending operations accidentally '
+        'recreating them without an expiry time. Defaults to 1 hour.',
```
hodgestar: Shouldn't the pending operations always create them with an expiring time too? Seems a bit scary to have any paths that can leave things in Redis forever? Perhaps I am just misunderstanding this help text though.

vumi/transports/smpp/smpp_transport.py
```diff
@@ -427,6 +437,7 @@ def process_submit_sm_event(self, message_id, event_type, remote_id,
                     "Could not retrieve failed message: %s" % (message_id,))
             else:
                 yield self.message_stash.delete_cached_message(message_id)
+                yield self.message_stash.expire_multipart_info(message_id)
```
hodgestar: I didn't see a test change for this failure path? I could just be sleepy though.

vumi/transports/smpp/smpp_transport.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, redis, config):
 
     def init_multipart_info(self, message_id, part_count):
         key = multipart_info_key(message_id)
-        expiry = self.config.third_party_id_expiry
+        expiry = self.config.submit_sm_expiry
```
hodgestar: Was this wrong originally?

hodgestar:
Left a couple of comments, otherwise looks good.

vumi/transports/smpp/config.py
jerith: Those are operations like "update this key in the hash", which we can't add expiration to in a single operation. We aren't expecting any of those after the message processing completes, but we can't guarantee that we won't get any. In the common case, adding an extra Redis call to every multipart `submit_sm_resp` we process doesn't seem worth it.

vumi/transports/smpp/smpp_transport.py
jerith: I didn't add one. I probably should.

vumi/transports/smpp/smpp_transport.py
jerith: Yes. That's why we have a whole week's worth instead of just a day's worth.

hodgestar:
:+1: once Travis is happy.

PR #969, opened 2015-09-08
Kill unnecessary SMPP protocols and factories

jerith:
We currently have six SMPP protocols and factories (two layers, three variants of each) and we only really need one.


jerith:
This is general cleanup work to make #965 less messy.

smn:
:+1:

PR #971, opened 2015-09-14
Move some SMPP from transport to service

jerith:
The `SmppService` is currently a very thin wrapper around a protocol. My plan for #965 is to move throttling out of the transport and into the service, which requires the service to play a more active role in message processing.


smn:
:+1: form me, nice to see the duplication removed from the processors. 
Where services fit in in Twisted's world of factories & protocols is still a bit of a mystery at times.

PR #972, opened 2015-09-15
Fix SMPP sequence number to message id TTL

jerith:
Like #968, but for the `sequence_number` key.

### DIFF
--- a/vumi/transports/smpp/protocol.py
+++ b/vumi/transports/smpp/protocol.py
@@ -332,6 +332,9 @@ class EsmeProtocol(Protocol):
         d.addCallback(
             self._handle_submit_sm_resp_callback, smpp_message_id,
             command_status, cb)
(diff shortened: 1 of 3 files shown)


jerith:
The ever-growing chain of callbacks on deletion is ugly, but this will be fixed in #965 and friends.

smn:
:+1:

smn:
At some point we're likely going to be able to do away with `third_party_id_expiry` entirely? We also probably want to trash the deprecated SMPP implementation at some point because that's going to be the only code still using that.

hodgestar:
:+1:

jerith:
> At some point we're likely going to be able to do away with `third_party_id_expiry` entirely?

Not if we want to be able to map delivery reports to vumi messages. The only thing still using it now is the `remote_id` to `message_id` mapping.

PR #973, opened 2015-09-21
Fix Travis pypy builds

jerith:
We occasionally hit a segfault bug in pypy<2.6.1, so we need to install our own never version.


hodgestar:
:+1: if Travis is happy.

PR #976, opened 2015-10-09
Don't set SMPP remote message ids for failure responses

justinvdm:
Currently, we store a mapping between remote message ids and our message ids in redis, and use this when we get a delivery report to publish a delivery report event with our message id. We don't need to do this for failure responses though, since they will never get a delivery report.


justinvdm:
Ready for review.

smn:
:+1:

smn:
:+1:

PR #977, opened 2015-10-09
Clear remote message redis keys when we get delivery reports

justinvdm:
Currently, we store a mapping between remote message ids and our message ids in redis, and use this when we get a delivery report to publish a delivery report event with our message id. Once we've published the delivery report event, we no longer need the remote message ids, yet we still keep them around for a week.

### DIFF
--- a/vumi/transports/smpp/smpp_transport.py
+++ b/vumi/transports/smpp/smpp_transport.py
@@ -236,6 +236,10 @@ class SmppMessageDataStash(object):
     def get_internal_message_id(self, smpp_message_id):
         return self.redis.get(remote_message_key(smpp_message_id))
 
+    def delete_remote_message_id(self, smpp_message_id):
+        key = remote_message_key(smpp_message_id)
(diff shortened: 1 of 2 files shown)


justinvdm:
Ready for review.

smn:
:+1: when travis

vumi/transports/smpp/smpp_transport.py
```diff
@@ -402,6 +406,8 @@ def handle_delivery_report(self, receipted_message_id, delivery_status):
         dr = yield self.publish_delivery_report(
             user_message_id=message_id,
             delivery_status=delivery_status)
+
+        yield self.message_stash.delete_remote_message_id(receipted_message_id)
```
hodgestar: I think there is a problem here -- if we delete the message when the delivery report status is something other than delivered, we won't be able to find the message if it does get delivered. Maybe we should only delete the remote message if the delivery report indicates complete failure or successful delivery?

PR #979, opened 2015-10-14
Infrastructure for publishing and consuming status events

justinvdm:
We need a way of publishing and consuming status event messages in workers.

The plan is to add two new connector types, one for publishing status event messages and one for consuming status event messages.

### DIFF
--- a/vumi/connectors.py
+++ b/vumi/connectors.py
@@ -1,8 +1,9 @@
 from twisted.internet.defer import gatherResults, inlineCallbacks, returnValue
 
 from vumi import log
-from vumi.message import TransportMessage, TransportEvent, TransportUserMessage
 from vumi.middleware import MiddlewareStack
+from vumi.message import (
+    TransportMessage, TransportEvent, TransportUserMessage, TransportStatus)
 
 
 class IgnoreMessage(Exception):
@@ -162,3 +163,28 @@ class ReceiveOutboundConnector(BaseConnector):
         return self.publish_event(TransportEvent(
             user_message_id=msg['message_id'], nack_reason=str(failure.value),
             event_type='nack'))
+
+
+class PublishStatusConnector(BaseConnector):
+    @inlineCallbacks
+    def setup(self):
+        yield self._setup_publisher('status')
+
+    def publish_status(self, msg, endpoint_name=None):
+        return self._publish_message('status', msg, endpoint_name)
+
+
+class ReceiveStatusConnector(BaseConnector):
+    @inlineCallbacks
+    def setup(self):
+        yield self._setup_consumer(
+            'status', TransportStatus, self.default_status_handler)
+
+    def default_status_handler(self, msg):
+        log.warning("No status handler for %r: %r" % (self.name, msg))
+
+    def set_status_handler(self, handler, endpoint_name=None):
+        self._set_endpoint_handler('status', handler, endpoint_name)
+
+    def set_default_status_handler(self, handler):
+        self._set_default_endpoint_handler('status', handler)
--- a/vumi/connectors.py
+++ b/vumi/connectors.py
@@ -170,8 +170,8 @@ class PublishStatusConnector(BaseConnector):
     def setup(self):
         yield self._setup_publisher('status')
 
-    def publish_status(self, msg, endpoint_name=None):
-        return self._publish_message('status', msg, endpoint_name)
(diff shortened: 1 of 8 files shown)


justinvdm:
Ready for review.

Planning on changing the base transport worker to have status event publishing capabilities in a separate PR.

vumi/message.py
```diff
@@ -429,3 +429,24 @@ def validate_fields(self):
             self.assert_field_present(extra_field)
             if not check(self[extra_field]):
                 raise InvalidMessageField(extra_field)
+
+
+class TransportStatus(TransportMessage):
+    """Message about a status event emitted by a transport.
+    """
```
justinvdm: It would be ideal if we could subclass `Message` instead (so that non-transport worker types could also publshing status events), but I think we need it to be subclass of `TransportMessage` for it to work with connectors.

vumi/connectors.py
```diff
@@ -162,3 +163,28 @@ def _ignore_message(self, failure, msg):
...
             event_type='nack'))
+
+
+class PublishStatusConnector(BaseConnector):
+    @inlineCallbacks
+    def setup(self):
+        yield self._setup_publisher('status')
+
+    def publish_status(self, msg, endpoint_name=None):
+        return self._publish_message('status', msg, endpoint_name)
```
hodgestar: Is there a point to allowing an endpoint to be set when calling `publish_status`?

vumi/connectors.py
```diff
@@ -162,3 +163,28 @@ def _ignore_message(self, failure, msg):
...
+            'status', TransportStatus, self.default_status_handler)
+
+    def default_status_handler(self, msg):
+        log.warning("No status handler for %r: %r" % (self.name, msg))
+
+    def set_status_handler(self, handler, endpoint_name=None):
+        self._set_endpoint_handler('status', handler, endpoint_name)
+
+    def set_default_status_handler(self, handler):
+        self._set_default_endpoint_handler('status', handler)
```
hodgestar: Same question about endpoints here -- perhaps we should only have `set_status_handler` for the moment?

vumi/tests/test_test_helpers.py
```diff
@@ -1231,6 +1333,37 @@ def test_dispatch_outbound_no_connector(self):
         self.assertEqual(
             broker.get_messages('vumi', 'fooconn.outbound'), [msg])
 
+    @inlineCallbacks
+    def test_dispatch_status(self):
+        """
+        WorkerHelper.dispatch_status() should dispatch an status message.
```
hodgestar: s/an/a/

vumi/tests/test_test_helpers.py
```diff
@@ -1528,6 +1661,25 @@ def test_make_reply(self):
...
+    @inlineCallbacks
+    def test_make_dispatch_status(self):
+        """
+        .make_dispatch_status() should build and dispatch a status message.
+        """
+        md_helper = MessageDispatchHelper(
+            MessageHelper(), WorkerHelper('fooconn'))
+        broker = md_helper.worker_helper.broker
+        broker.exchange_declare('vumi', 'direct')
+        self.assertEqual(broker.get_messages('vumi', 'fooconn.outbound'), [])
```
hodgestar: s/fooconn.outbound/fooconn.status/

vumi/message.py
hodgestar: Correct.

hodgestar:
Some minor comments, otherwise looks good.

vumi/connectors.py
justinvdm: Probably not, will remove.

hodgestar:
I have a very bikesheddy comment -- would `healthy connector` be a better name than `status connector`? Trying to decide if it feels better enough to warrant running around and changing?

justinvdm:
> I have a very bikesheddy comment -- would healthy connector be a better name than status connector? Trying to decide if it feels better enough to warrant running around and changing?

Thought about this too, had a vague preference for 'health' over 'status' in general, but we already seem to refer to it as 'status' in junebug (e.g. `status_url`), so I think consistency would be good.

hodgestar:
Cool. Let's keep it as it is.

justinvdm:
Ready for re-review.

hodgestar:
:+1:

PR #982, opened 2015-10-15
Enable status publishing via config flag

justinvdm:
The plan is to disable actual publishing of statuses by `publish_status()` by default.


justinvdm:
Ready for review.

hodgestar:
:+1:

PR #983, opened 2015-10-15
Publish status event on connect for SMPP transport

justinvdm:


vumi/transports/smpp/smpp_transport.py
```diff
@@ -298,6 +298,15 @@ def _reject_for_invalid_address(self, message, field):
         return self.publish_nack(
             message['message_id'], u'Invalid %s: %s' % (field, message[field]))
 
+    def on_connection(self):
+        return self.publish_unbound()
+
+    def publish_unbound(self):
+        return self.publish_status('smpp', 'major', reasons=[{
+            'type': 'unbound',
+            'message': 'Connected but not bound',
```
justinvdm: Not too sure on this type and message.

justinvdm:
Ready for review.

justinvdm:
Sorry, seeing some weird test failures, fixing, probably not ready review.

justinvdm:
Actually ready for review now.

vumi/transports/smpp/smpp_transport.py
```diff
@@ -298,6 +298,15 @@ def _reject_for_invalid_address(self, message, field):
...
             message['message_id'], u'Invalid %s: %s' % (field, message[field]))
 
+    def on_connection(self):
+        return self.publish_status_unbound()
+
+    def publish_status_unbound(self):
+        return self.publish_status('smpp', 'major', reasons=[{
+            'type': 'unbound',
+            'message': 'Connected but not bound',
+        }])
```
justinvdm: Wondering if there are any `details` we should add. Thought about adding the host and port details, but that is a thing that should already be known by the thing that created the transport. In Junebug's case, the client using the Junebug API should know host and port details since it is the thing that configured the transport to use those details.

rudigiesler: I think what I noted here: https://github.com/praekelt/vumi/pull/984#discussion_r42141800 , applies here too. We don't really have any freely available information besides the config information.

justinvdm: +1, thanks.

rudigiesler:
:+1:

PR #984, opened 2015-10-15
Publish status event on bind for SMPP transport

justinvdm:


justinvdm:
Ready for review.

vumi/transports/smpp/smpp_transport.py
```diff
@@ -299,6 +299,16 @@ def _reject_for_invalid_address(self, message, field):
...
 
     @inlineCallbacks
+    def on_smpp_bind(self):
+        yield self.publish_status_bound()
+
+    def publish_status_bound(self):
+        return self.publish_status('smpp', 'good', reasons=[{
+            'type': 'bound',
+            'message': 'Bound',
+        }])
```
justinvdm: Ideas on any `details` we should add here?

rudigiesler: I don't think we have any information at this point that would be useful? I think things like the time it took to bind, or smpp version. But I don't think that any of those things are easily available here, we would have to implement them separately. I think for now we can leave details empty, and if we find that some things are useful to know, then we can implement them.

rudigiesler:
Do you think we should be adding the relevant documentation with each of these PRs, or just do one big PR in the end with all of these changes?

vumi/transports/smpp/smpp_transport.py
justinvdm: +1

justinvdm:
> Do you think we should be adding the relevant documentation with each of these PRs, or just do one big PR in the end with all of these changes?

Was thinking about this, decided to go for big docs PR at the end of all these changes to avoid needing to keep the docs up to date on each small change we make to these first few events.

rudigiesler:
:+1:

PR #990, opened 2015-10-19
Rename 'unbound' status event to 'binding' for SMPP

justinvdm:
`unbound` doesn't imply we are attempting to bind, `binding` does.


justinvdm:
Ready for review.

rudigiesler:
:+1: when travis is happy

PR #978, opened 2015-10-09
Only remove remote ids for failure or success delivery reports

justinvdm:
Currently we also remove for pending delivery reports.

PR #991, opened 2015-10-19
Don't remove remote message ids for SMPP delivery reports

justinvdm:
Since #977 and #978, we remove message ids for failure and success delivery reports. It is common for us to get multiple delivery reports for the same message though, so we can't really do this.

### DIFF
--- a/vumi/transports/smpp/tests/test_smpp_transport.py
+++ b/vumi/transports/smpp/tests/test_smpp_transport.py
@@ -1523,7 +1523,7 @@ class SmppTransceiverTransportTestCase(SmppTransportTestCase):
 
         self.assertTrue(
             remote_id_ttl > 23,
-            "remote_id_ttl (%s) > final_dr_third_party_id_expiry (23)"
+            "remote_id_ttl (%s) <= final_dr_third_party_id_expiry (23)"
             % (remote_id_ttl,))
 
         pdu = DeliverSM(sequence_number=1, esm_class=4)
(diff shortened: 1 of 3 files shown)


justinvdm:
Ready for review.

rudigiesler:
:+1: from me, assuming that people are happy with the 1 hour timeout

vumi/transports/smpp/tests/test_smpp_transport.py
```diff
@@ -1524,14 +1530,20 @@ def test_delivery_report_delivered_delete_stored_remote_id(self):
 
         yield self.tx_helper.wait_for_dispatched_events(1)
 
-        self.assertFalse(
-            (yield transport.redis.exists(remote_message_key('foo'))))
+        self.assertEqual(
+            (yield transport.redis.ttl(remote_message_key('foo'))), 23)
```
jerith: We probably want `<=` rather than `==` here, otherwise timing issues (such as a heavily loaded build machine) could cause spurious failures.

jerith:
Looks good, except for the assertion comment. You can look at other SMPP tests that make assertions about Redis TTLs to see how they handle the problem.

justinvdm:
Ready for re-review.

hodgestar:
One minor comment about assertion message text, otherwise looks good to me.

justinvdm:
Assertion messages fixed.

hodgestar:
:+1:

PR #992, opened 2015-10-19
Rename status event level names

justinvdm:
At the moment we use `good`, `minor` and `major`. The plan is to change these to `ok`, `degraded` and `down`, since those are better descriptions for what the status events mean.

### DIFF
--- a/vumi/transports/tests/test_base.py
+++ b/vumi/transports/tests/test_base.py
@@ -123,12 +123,12 @@ class TestBaseTransport(VumiTestCase):
         })
 
         msg = yield transport.publish_status(
-            status='major',
+            status='down',
             component='foo',
             type='bar',
             message='baz')
 
-        self.assertEqual(msg['status'], 'major')
+        self.assertEqual(msg['status'], 'down')
         self.assertEqual(msg['component'], 'foo')
         self.assertEqual(msg['type'], 'bar')
         self.assertEqual(msg['message'], 'baz')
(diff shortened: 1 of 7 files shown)


justinvdm:
Ready for review.

jerith:
I'd be happier if these were named constants, so it's harder to accidentally misspell them or something.

hodgestar:
@jerith Doesn't message validation already check that?

jerith:
Ah, I didn't see that. Thanks. :-)

I still think it makes more sense to use something more "enumeration-like" than arbitrary strings given that there is a fixed set of valid values, but I'm also okay with strings+validation.

justinvdm:
@hodgestar Hmm, forgot that we do validation, good point.

Changed to using class-level variables, but happy to revert the change.

hodgestar:
I'm okay with keeping the constants. They're sometimes a little annoying in Python because one ends up importing weird modules and classes just to access one string, but in Vumi itself that's not really an issue.

hodgestar:
And we have validation too, so woot.

vumi/transports/tests/test_base.py
```diff
@@ -123,12 +123,12 @@ def test_publish_status(self):
         })
 
         msg = yield transport.publish_status(
-            status='major',
+            status='down',
```
hodgestar: Do we want to use `TransportStatus.STATUS_DOWN`, etc in `test_base.py` too?

hodgestar:
Other than the final tiny comment about a missed opportunity to used `TransportStatus.STATUS_X`, looks good to me.

justinvdm:
Decided to revert the class-level variables approach, agree with [this comment](https://github.com/praekelt/vumi/pull/992#issuecomment-149237309).

Ready for re-review.

hodgestar:
:+1:

PR #996, opened 2015-10-21
Ensure type and message fields are present for TransportStatus

justinvdm:
These fields should be mandatory, but we don't ensure they are present when validating a `TransportStatus` message.


justinvdm:
Ready for review.

rudigiesler:
:+1:

PR #997, opened 2015-10-22
Remove unused SimpleDispatcher

jerith:
It hasn't been touched in several years and it appears to be the only thing that uses publishers with arbitrary routing keys.


hodgestar:
Looks good. There are also a reference to `SimpleDispatcher` in `etc/supervisord.smpp_test.conf` though.

hodgestar:
Do we want to fix `config/smpp_test.yaml` and `config/smpp_test_http_post.yaml` so they have the same `transport _name` now that the dispatcher is gone?

jerith:
No, because they're both transports. I think the stuff I removed was broken anyway.

hodgestar:
Oh. :/

hodgestar:
:+1:

PR #998, opened 2015-10-22
Split off a separate DynamicPublisher and streamline it

jerith:


smn:
:+1: wow, this ancient low level stuff is getting attention. \o/

PR #974, opened 2015-09-22
Fake connections for tests

jerith:
Using real network connections (even over localhost) in tests is unnecessary in most cases and leads to occasional random build failures.

This branch uses the same ideas as the recently-added `FakeSMSC` to test various other transports, applications, and utilities.

[Review on Reviewable](https://reviewable.io/reviews/praekelt/vumi/974)

PR #1001, opened 2015-10-30
Fix StreamingClient

jerith:
#974 changed `vumi.transports.vumi_bridge.client.StreamingClient`'s constructor in a way that breaks vumi-go's HTTP API application worker.


hodgestar:
:+1:

This is only for the deprecated streaming API, correct?

jerith:
Yes, specifically its tests.

PR #1003, opened 2015-11-03
More robust data collection in MTN NG tests

jerith:
The current `WaitForDataMixin` can miss data in certain circumstances. Using a `DeferredQueue` instead is both simpler and more robust.


JayH5:
:+1:

PR #1004, opened 2015-11-04
Add TransportStatus tracker

rudigiesler:
We need a tool to make sure that we don't send the same TransportStatus repeatedly, we only want to send a status event when the status changes.

### DIFF
--- a/vumi/utils.py
+++ b/vumi/utils.py
@@ -452,3 +452,54 @@ def safe_routing_key(routing_key):
 
 def generate_worker_id(system_id, worker_id):
     return "%s:%s" % (system_id, worker_id,)
+
+
+class StatusEdgeDetector(object):
+    '''Assists with finding if a TransportStatus is a change in the status,
+    compared to previous statuses, or just a repeat. Will be useful to only
+    publish statuses on state change.'''
+
+    def __init__(self):
+        self.state = dict()
+        self.types = dict()
+
+    def check_status(self, status):
+        '''
+        Checks to see if the current status is a repeat. If it is, None is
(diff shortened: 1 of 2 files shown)


vumi/utils.py
```diff
@@ -452,3 +452,54 @@ def safe_routing_key(routing_key):
...
+
+    def _check_state(self, status, component):
+        state = self._get_state(component)
+        if state != status:
+            self._remove_types(component)
+            self._set_state(component, status)
+
+    def _check_type(self, type, component):
+        types = self._get_types(component)
+        if type not in types:
```
justinvdm: A little bikesheddy, but could we avoid the `type` reserved word?

vumi/utils.py
```diff
@@ -452,3 +452,54 @@ def safe_routing_key(routing_key):
...
+            self._add_type(component, type)
+            return True
```
justinvdm: Should we maybe return `False` otherwise?

justinvdm:
Left two minor comments, looks good.

rudigiesler:
Comments addressed

vumi/utils.py
```diff
@@ -452,3 +452,55 @@ def safe_routing_key(routing_key):
...
+    def _set_state(self, component, state):
+        self.state[component] = state
+
+    def _get_types(self, component):
+        return self.types.get(component, set())
+
+    def _add_type(self, component, type):
+        if component not in self.types:
+            self.types[component] = set()
+        self.types[component].add(type)
```
justinvdm: Could we avoid using `type` here as well?

justinvdm:
:+1:

PR #1005, opened 2015-11-04
Add a helper method for publishing status events in http rpc transport

rudigiesler:
We need to add a helper method to the base http rpc transport that makes use of the component added in #1004.

### DIFF
--- a/vumi/transports/httprpc/httprpc.py
+++ b/vumi/transports/httprpc/httprpc.py
@@ -200,8 +200,7 @@ class HttpRpcTransport(Transport):
     def add_status(self, **kw):
         '''Publishes a status if it is not a repeat of the previously
         published status.'''
-        status = TransportStatus(**kw)
-        if self.status_detect.check_status(status):
+        if self.status_detect.check_status(**kw):
             return self.publish_status(**kw)
         return succeed(None)
 
--- a/vumi/transports/httprpc/httprpc.py
+++ b/vumi/transports/httprpc/httprpc.py
@@ -3,7 +3,7 @@
 import json
 
 from twisted.cred.portal import Portal
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import inlineCallbacks, succeed
 from twisted.internet import reactor
 from twisted.internet.task import LoopingCall
 from twisted.web import http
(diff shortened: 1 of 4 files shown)


rudigiesler:
Ready for review.

vumi/transports/httprpc/httprpc.py
```diff
@@ -193,6 +195,16 @@ def setup_transport(self):
             ],
             self.web_port)
 
+        self.status_detect = StatusEdgeDetector()
+
+    def add_status(self, **kw):
+        '''Publishes a status if it is not a repeat of the previously
+        published status.'''
+        status = TransportStatus(**kw)
+        if self.status_detect.check_status(status):
```
hodgestar: Given how `publish_status` accepts `**kw`, I wonder if the edge detector shouldn't work the same way? Or at least have the option of being used that way.

vumi/transports/httprpc/httprpc.py
```diff
@@ -193,6 +195,16 @@ def setup_transport(self):
...
+            return self.publish_status(**kw)
+        return succeed(None)
```
hodgestar: Is return a deferred here just future proofing?

hodgestar:
Left two comments, but looks good.

hodgestar:
On other thought -- does `add_status` belong on the base transport perhaps?

vumi/transports/httprpc/httprpc.py
rudigiesler: Yes I didn't realise that `publish_status` worked that way, I thought that it accepted a `TransportStatus`. I'll change the edge detector to work that way.

vumi/transports/httprpc/httprpc.py
rudigiesler: It's not strictly necessary, I just think it's nice to always return a deferred.

rudigiesler:
I don't know if it belongs in the base transport, since not all transports need it. Things like the SMPP transport only sends a status on certain events, so it won't have the problem of sending the same status over and over again. It could maybe go in the base transport if we find that other transports need it, but so far it's just the HTTP RPC transport that needs it

vumi/transports/httprpc/httprpc.py
hodgestar: Agreed.

rudigiesler:
Changed the way that the edge detector takes in arguments. Ready for review again.

justinvdm:
:+1: from me.

PR #1006, opened 2015-11-04
Remove get_stubbed_channel

jerith:
`get_stubbed_channel()` is an old test helper that relies on implementation details of `fake_amqp` that are changing soon. It's easily replaced (in the few tests that use it) by a couple of lines that construct an actual channel from a fake client.

### DIFF
--- a/vumi/blinkenlights/metrics.py
+++ b/vumi/blinkenlights/metrics.py
@@ -75,7 +75,8 @@ class MetricManager(object):
         Stop the metric polling and publishing task.
         """
         if self._task:
-            self._task.stop()
+            if self._task.running:
+                self._task.stop()
             self._task = None
 
     def publish_metrics(self):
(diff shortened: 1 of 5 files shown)


vumi/blinkenlights/metrics.py
```diff
@@ -75,7 +75,8 @@ def stop_polling(self):
         Stop the metric polling and publishing task.
         """
         if self._task:
-            self._task.stop()
+            if self._task.running:
+                self._task.stop()
```
hodgestar: Seems a bit unrelated to this issue, although seems like a good idea? Just checking it didn't sneak in unintentionally.

hodgestar:
One question, but otherwise :+1:.

vumi/blinkenlights/metrics.py
jerith: One of the tests now tries to stop a task that was never started (because we're adding cleanup that wasn't there before) and that broke the manager teardown.

hodgestar:
:+1:

PR #1007, opened 2015-11-05
Add status for dmark transport requests

rudigiesler:


rudigiesler:
Ready for review.

vumi/transports/dmark/dmark_ussd.py
```diff
@@ -113,8 +113,20 @@ def handle_raw_inbound_message(self, request_id, request):
             log.msg('Unhappy incoming message: %r' % (errors,))
             self.finish_request(
                 request_id, json.dumps(errors), code=http.BAD_REQUEST)
+            yield self.add_status(
+                component='request',
```
justinvdm: Can't remember if this is a thing that we discussed. Did we decide on separate request and response components? Wondering if it would make more sense to have an `http` component. Thoughts?

vumi/transports/dmark/dmark_ussd.py
```diff
@@ -113,8 +113,20 @@ def handle_raw_inbound_message(self, request_id, request):
...
+                type='invalid_inbound_fields',
+                message='Invalid inbound fields',
+                details=errors)
             return
 
+        yield self.add_status(
+            component='request',
+            status='ok',
+            type='request_decoded',
+            message='Request decoded',)
```
justinvdm: A little bikesheddy, but wondering if we should use 'parsed' instead of 'decoded', 'decoded' might imply character encoding or data deserializing which both seem like subsets of parsing.

justinvdm:
Two minor comments, looks good :)

vumi/transports/dmark/dmark_ussd.py
rudigiesler: I think we decided on making it separate, in case a request comes in that's okay (we send an ok status), then if we timeout sending a response (we send a down status), and it would toggle between sending those two statuses for every message.

vumi/transports/dmark/dmark_ussd.py
rudigiesler: Sure, parsed sounds better.

vumi/transports/dmark/dmark_ussd.py
justinvdm: Ah, I see, ok cool.

vumi/transports/dmark/dmark_ussd.py
```diff
@@ -108,13 +108,48 @@ def session_event_for_transaction(self, transaction_id):
...
+            values, errors = self.get_field_values(request, self.EXPECTED_FIELDS)
+        except UnicodeDecodeError:
+            log.msg('Bad request encoding: %r' % request)
+            request_dict = {
+                'uri': request.uri,
+                'method': request.method,
+                'path': request.path,
+                'content': request.content.read(),
+                'headers': dict(request.requestHeaders.getAllRawHeaders()),
+            }
```
justinvdm: Might be worth putting this dictionaryifying into a utility function if we start using it in other transports. Not something for now though.

vumi/transports/dmark/tests/test_dmark_ussd.py
```diff
@@ -102,6 +105,43 @@ def test_inbound_begin(self):
...
+    @inlineCallbacks
+    def test_inbound_cannot_decode(self):
+        '''If the content cannot be decoded, an error shoould be sent back'''
+        user_content = "Who are you?".encode('utf-32')
+        response = yield self.tx_helper.mk_request(ussdRequestString=user_content)
+        self.assertEqual(response.code, 400)
+
+        body = json.loads(response.delivered_body)
+        request = body['invalid_request']
+        self.assertEqual(request['method'], 'GET')
```
justinvdm: Would it maybe be worth checking the other request data?

vumi/transports/dmark/tests/test_dmark_ussd.py
```diff
@@ -102,6 +105,43 @@ def test_inbound_begin(self):
...
+        user_content = "Who are you?".encode('utf-32')
+        response = yield self.tx_helper.mk_request(ussdRequestString=user_content)
+
+        [status] = self.tx_helper.get_dispatched_statuses()
+        self.assertEqual(status['component'], 'request')
+        self.assertEqual(status['status'], 'down')
+        self.assertEqual(status['type'], 'invalid_encoding')
+        self.assertEqual(status['message'], 'Invalid encoding')
+
+
```
justinvdm: One too many newlines, I think pep8 would complain.

vumi/transports/dmark/tests/test_dmark_ussd.py
```diff
@@ -102,6 +105,43 @@ def test_inbound_begin(self):
...
+    def test_inbound_cannot_decode_status(self):
+        '''If the request cannot be decoded, a status event should be sent'''
+        user_content = "Who are you?".encode('utf-32')
+        response = yield self.tx_helper.mk_request(ussdRequestString=user_content)
+
+        [status] = self.tx_helper.get_dispatched_statuses()
+        self.assertEqual(status['component'], 'request')
+        self.assertEqual(status['status'], 'down')
+        self.assertEqual(status['type'], 'invalid_encoding')
+        self.assertEqual(status['message'], 'Invalid encoding')
```
justinvdm: Should we check the status message's details?

justinvdm:
Hmm, just realised I probably started commenting again too earlier, sorry about that, will wait for the PR to be ready for re-review.

vumi/transports/dmark/tests/test_dmark_ussd.py
rudigiesler: So there's a small issue in why I didn't do that. It seems that the URI is partially decoded (`%20` turned into `+`). I have no idea where this happens, or how to stop it, but it does mean that I cannot decode the URL and check it.

rudigiesler:
Should be ready for review.

justinvdm:
:+1:

PR #1008, opened 2015-11-05
WorkerHelper should support getting messages after pending deliveries

jerith:
At the moment, we can either get current messages or wait for a specific number of messages. It would be useful to be able to wait for any pending deliveries to complete and then get whatever messages are available. (This is particularly helpful for #1000, which breaks the assumption that the client can wait for the broker to receive published messages.)

In order to maintain compatibility while avoiding API bloat, I'm making the `amount` parameter optional in the various `WorkerHelper.wait_for_dispatched_*` methods and treating a value of `None` (the default) as an indication that we should wait for pending deliveries to finish instead of waiting for a particular number of messages.

### DIFF
--- a/vumi/blinkenlights/tests/test_metrics.py
+++ b/vumi/blinkenlights/tests/test_metrics.py
@@ -25,8 +25,9 @@ class TestMetricPublisher(VumiTestCase):
         reactor.callLater(delay, lambda: d.callback(None))
         return d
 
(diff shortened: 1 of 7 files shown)


jerith:
The core change is done, but I also want to update a bunch of tests to use the new functionality as part of this PR.

smn:
:+1:

PR #1009, opened 2015-11-05
Add status for dmark transport timeouts

rudigiesler:


rudigiesler:
Ready for review.

vumi/transports/dmark/dmark_ussd.py
```diff
@@ -191,12 +198,54 @@ def handle_outbound_message(self, message):
...
+                    })
+            else:
+                return self.add_status(
+                    component='response',
+                    status='ok',
+                    type='response_sent',
+                    message='Response sent',
+                    details={
+                        'response_time': response_time,
+                    })
```
justinvdm: Two thoughts:
  - it would be quite useful if we could have these response times be configurable. If we could do something like `response_intervals: [10, 60]` (so the interval amounts are configurable, and the number of intervals are configurable), I think that would be ideal.
  - should some of the logic for this maybe live in the base rpc transport? Maybe we could hook into it via some callback method with a signature like `def on_response_interval(interval):` or something, and do the `self.add_status` there?

Thoughts?

vumi/transports/dmark/tests/test_dmark_ussd.py
```diff
@@ -275,3 +275,65 @@ def test_nack_http_http_response_failure(self):
...
+    def test_status_degraded_slow_response(self):
+        '''A degraded status event should be sent if the response took longer
+        than 1 second.'''
+        d = self.tx_helper.mk_request()
+        [msg] = yield self.tx_helper.wait_for_dispatched_inbound(1)
+        yield self.tx_helper.clear_dispatched_statuses()
+
+        # 1.5 second reply
+        timestamp = self.transport.session_timestamps[msg['message_id']]
+        self.transport.session_timestamps[msg['message_id']] = timestamp - 1.5
```
justinvdm: Would it maybe be a bit more thorough to advance the clock to test this?

vumi/transports/dmark/tests/test_dmark_ussd.py
```diff
@@ -275,3 +275,65 @@ def test_nack_http_http_response_failure(self):
...
+        self.tx_helper.dispatch_outbound(msg.reply('foo'))
+        response = yield d
+
+        [status] = yield self.tx_helper.get_dispatched_statuses()
+        self.assertEqual(status['status'], 'down')
+        self.assertEqual(
+            status['reasons'], ['Response took longer than 10 seconds'])
+        self.assertEqual(status['component'], 'response')
+        self.assertEqual(status['type'], 'slow_response')
+        self.assertEqual(status['message'], 'Very slow response')
```
justinvdm: Do we maybe want a test for when we send an outbound message but there is no corresponding message id in `session_timestamps`? Thinking of cases where we restart the transport.

justinvdm:
Not too important idea: response times are something well suited to middleware, wondering if there is a way we could have a middleware for this instead? Thoughts?

justinvdm:
> Not too important idea: response times are something well suited to middleware, wondering if there is a way we could have a middleware for this instead? Thoughts?
Actually, nvm, what we are trying to do is way too tied into the requests and responses that only the transport knows about to have a middleware do this.

justinvdm:
Left a few comments, looks good.

vumi/transports/dmark/dmark_ussd.py
rudigiesler: We could do that. We could also maybe have the `set_request_start` and `set_request_end` functions in the base transport. We could even have it all called automatically, and have each transport just configure the response times.

The only problem which what you're listing there for the configuration, is when do we stop making degraded status and start making down statuses? And what's the value in having multiple intervals if we only have two status types?

justinvdm: > The only problem which what you're listing there for the configuration, is when do we stop making degraded status and start making down statuses? And what's the value in having multiple intervals if we only have two status types?

Hmm, true, maybe we should restrict it to two configurable intervals, `response_time_degraded` and `response_time_down` intervals, instead of a variable number of intervals?

justinvdm: > We could do that. We could also maybe have the set_request_start and set_request_end functions in the base transport. We could even have it all called automatically, and have each transport just configure the response times.

I have a vague preference for the `self.add_status` calls happening in the transport itself instead of in the base rpc class, seems like something we might want to control per-transport (maybe some transports have special cases, or something). Don't mind us trying the completely automatic approach though.

rudigiesler:
@justinvdm: Addressed your comments. Ready for review again.

justinvdm:
:+1:

PR #1011, opened 2015-11-06
Change response time to base transport calling subclass

rudigiesler:
Currently, the subclass for the http transport needs to call `set_request_start` and `set_request_end` in the base function to be able to send status events for long response times. Change this to have the base httprpc transport calculate the times, and if they exceed the configured values, call a function in the subclass transport that has the logic to do the correct thing (for dmark, send the relevant status event).

### DIFF
--- a/vumi/transports/dmark/dmark_ussd.py
+++ b/vumi/transports/dmark/dmark_ussd.py
@@ -147,8 +147,6 @@ class DmarkUssdTransport(HttpRpcTransport):
                 details=errors)
             return
 
-        yield self.set_request_start(request_id)
-
         yield self.add_status(
             component='request',
             status='ok',
@@ -198,8 +196,6 @@ class DmarkUssdTransport(HttpRpcTransport):
         response_id = self.finish_request(
             message['in_reply_to'], json.dumps(response_data))
 
-        yield self.set_request_end(message['in_reply_to'])
-
         if response_id is not None:
             ack = yield self.publish_ack(
                 user_message_id=message['message_id'],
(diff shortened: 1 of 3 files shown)


rudigiesler:
Ready for review.

vumi/transports/dmark/dmark_ussd.py
```diff
@@ -211,3 +207,53 @@ def handle_outbound_message(self, message):
...
+
+    def on_degraded_response_time(self, message_id, time):
+        request = self.get_request(message_id)
+        # We send different status events for error responses
+        if request.code < 200 or request.code >= 300:
+            return
+        return self.add_status(
+            component='response',
+            status='degraded',
+            type='slow_response',
```
justinvdm: Wondering if this should have a different `type` to the down status event? They seem like different types.

vumi/transports/dmark/dmark_ussd.py
```diff
@@ -211,3 +207,53 @@ def handle_outbound_message(self, message):
...
+            ],
+            details={
+                'response_time': time,
+            })
+
+    def on_good_response_time(self, message_id, time):
+        request = self.get_request(message_id)
+        # We send different status events for error responses
+        if request.code < 200 or request.code >= 300:
+            return
```
justinvdm: Sorry for not catching this case in review. I think the `3xx` range might be okay, since those are for redirections. Maybe this should be `> 399`?

justinvdm: Separate thing -- could we add tests for these cases?

justinvdm:
Left some minor comments, looks good :)

rudigiesler:
@justinvdm Addressed your comments. Ready for another review.

justinvdm:
:+1: when travis is happy

PR #1013, opened 2015-11-11
Remove piles of long-deprecated test helper code

jerith:
This branch removes a whole lot of stuff that other projects may still use (although I did check a few of our big repos and only found one thing that needed fixing), so it's probably worth bumping the next release version to 0.6.0 or something.

I'm expecting to do more things like this in the near future, so it may be best to start a separate 0.6.0 release branch to collect all of them so we don't end up with a series of potentially breaking changes?


hodgestar:
:+1:

PR #1014, opened 2015-11-13
Fix publisher exchange setup

jerith:
When we changed how dynamic publishers work, we somehow lost the `durable=True` that was applied to the exchange creation. This wasn't caught in tests because FakeAMQP ignores the `durable` param on exchanges.


hodgestar:
:+1: when Travis is happy.

PR #1017, opened 2015-11-18
Add support for event exports.

hodgestar:
After some discussion the consensus appears to be something along the lines of:

* Being able to export events would be good.
* Exporting should be provided by a separate event exporting resource (to go with the ones for inbound and outbound)
* We should not retrieve events by looking up the events for particular outbound messages, because this results in a Riak *index* query per outbound message (which is very bad for performance).
* This leaves us retrieving events by event timestamp (not the most terrible thing).


hodgestar:
Ready for review.

Aside: The outbound and inbound message exporting has a lot of tests. I added fewer for the events because all of the range-parsing code being tested is shared between the three implementations.

jerith:
What's the reason for adding the `keys_only` param to the existing methods?

hodgestar:
`keys_only` option removed. It was there to allow `list(page)` to return a list of keys, but I've worked around that.

hodgestar:
Merged in develop to fix test that was fixed in develop.

Ready for re-review.

hodgestar:
Build failures are only on Twisted 13 and seem unrelated.

vumi/components/tests/test_message_store.py
```diff
@@ -112,6 +112,27 @@ def create_outbound_messages(self, batch_id, count, start_timestamp=None,
             messages.append(msg)
         returnValue(messages)
 
+    def _create_event(self, event_type, timestamp):
+        assert event_type in ('ack', 'nack', 'delivery_report')
+        maker = getattr(self.msg_helper, 'make_%s' % (event_type,))
+        return maker(timestamp=timestamp)
```
jerith: Can we use `self.msg_helper.make_event()` here instead of `getattr()`?

hodgestar: I did try that but we end up needing to duplicate all of the default logic
from the individual `make_ack`, `make_nack` & `make_delivery_report`
methods.

jerith: Hrm. Maybe dict-based dispatch (instead of the assert and `getattr`), then? I'm not enthusiastic about using `getattr` for this stuff.

hodgestar: Works for me. Will implement.

hodgestar:
Ready for re-re-review. :)

jerith:
:+1: when Travis is finished.

PR #1018, opened 2015-11-24
Update pypy build to 4.0.1

jerith:


JayH5:
Personally, I would prefer if `RIAK_VERSION` and `TWISTED_VERSION` were explicitly set for all the builds in the matrix so that it is clear exactly what is to be run.

jerith:
Having a long list of envvars on every build matrix entry can make it harder to see which build have non-default values, which is usually more important than knowing the values for every build. Also, the Travis UI displays everything from the `env` section of the build matrix, which can make it hard to differentiate between builds when the list is long.

I should explicitly set the default values in the `before_install` section instead of having that happen in the places the envvars are used, though.

JayH5:
:+1: when travis

jerith:
The build failures above are actually build cancellations. We have a lot of stuff in the queue and the earlier builds aren't really useful anymore.

PR #1024, opened 2016-02-23
Feature/we chat transport integration

codiebeulaine:
@hodgestar just adding some comments to docstring about how it works?


justinvdm:
:+1:

PR #1026, opened 2016-03-03
Add response error constant to Redis managers so that exceptions can be caught sanely.

hodgestar:
Currently one has to do:
```python
try:
    redis.op(...)
except Exception:
    ...
```
for most redis operations because the class of exception raised depends on the kind of redis one has (sync, async, or fake).

Adding a `RESPONSE_ERROR` attribute to the redis manager would allow us to do:
```
try:
    redis.op(...)
except redis.RESPONSE_ERROR:
    ...
```
Much nicer.

### DIFF
--- a/vumi/persist/redis_manager.py
+++ b/vumi/persist/redis_manager.py
@@ -1,9 +1,11 @@
 # -*- test-case-name: vumi.persist.tests.test_redis_manager -*-
 
 import redis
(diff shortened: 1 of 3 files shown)


hodgestar:
@KaitCrawford Ready for review.

KaitCrawford:
:+1: looks good to me

PR #1028, opened 2016-03-22
Cleanup Vumi bridge and add Junebug support

KaitCrawford:


vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
```diff
@@ -65,7 +64,7 @@ def get_next_request(self):
 
 class TestGoConversationTransport(TestGoConversationTransportBase):
 
-    transport_class = GoConversationClientTransport
+    transport_class = GoConversationTransport
```
hodgestar: We need to delete these old client transport tests. They're not failing currently because they're accidentally not being run (because the both this class and the server test class are now called `TestGoConversationTransport`).

vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
```diff
@@ -183,17 +182,15 @@ def test_reconnecting(self):
...
 
     def test_server_settings_without_configs(self):
         return self.assertFailure(self.get_transport(), ConfigError)
 
     def get_configured_transport(self):
-        return self.get_transport(
-            message_path='messages.json', event_path='events.json',
-            web_port='0')
+        return self.get_transport(web_path='', web_port='0')
```
hodgestar: We should probably set `web_path` to something that is not `''` to demonstrate that it's working.

hodgestar:
Left two small comments, and we need to look at why tests are failing, but otherwise looks good.

hodgestar:
Given that unrelated middleware tests are failing for all builds except the one pinned to Twisted 13.2, I suspect this indicates more tests broken by Twisted 16.0.

rudigiesler:
@hodgestar Looking at the middleware tests that are failing, looks like there's a twisted import (to do with ssh), that if it fails, the tests aren't run. So it's very possible that the latest twisted fixed that module import, and that test has been broken for a very long time. (Either way, completed unrelated to this PR).

hodgestar:
Creating #1029 to deal with the manhole middleware test failures.

KaitCrawford:
@hodgestar @rudigiesler Please give this another look.
I'm not sure the last commit is what you meant hodgestar, I might be misunderstanding how web_path gets used.
I also think I should probably be updating the status in more places but I'm struggling to identify where.

hodgestar:
@KaitCrawford Could you merge develop into this branch so we can check if tests pass?

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -152,101 +142,23 @@ def get_auth_headers(self):
...
-    def disconnect_api_clients(self):
-        self.message_client.disconnect()
-        self.event_client.disconnect()
-
+    def update_status(self, **kw):
+        '''Publishes a status if it is not a repeat of the previously
+        published status.'''
+        if self.status_detect.check_status(**kw):
+            yield self.publish_status(**kw)
+            # TODO: Notify Junebug
```
hodgestar: The line above notifies Junebug, so we can remove this TODO. :)

hodgestar:
Other than removing the TODO, looks good to me!

KaitCrawford:
Really? \o/ whoohooo!

rudigiesler:
Looks good to me too. :+1:

hodgestar:
Oh, I realized something -- do we want to add status events for inbound messages too? Just "good" if we receive a valid request and "bad" if we receive a broken request?

hodgestar:
I also remembered that we wanted to only publish `acks` once we received an `ack` from Vumi Go (so publish a `nack` if submitting the message to Vumi Go fails and then publish whatever event Vumi Go gives us).

hodgestar:
Maybe this means we should also re-think the components a bit. Maybe we want separate components for `http-requests-to-vumi-go`, `http-requests-from-vumi-go` and `events-from-vumi-go` (except with better names)?

KaitCrawford:
@hodgestar the names are probably still horrible :)

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -152,101 +142,22 @@ def get_auth_headers(self):
...
+        '''Publishes a status if it is not a repeat of the previously
+        published status.'''
+        if self.status_detect.check_status(**kw):
+            yield self.publish_status(**kw)
 
-class GoConversationTransport(GoConversationClientTransport):
+    def add_status_bad_req(self, component):
+        return self.update_status(
+            status='down', component=component, type='bad_request',
+            message='Bad request received')
```
hodgestar: We need to set `type` and `message` more appropriately for the different error / success cases (e.g. `Bad request received` is not correct when we're failing to make requests).

Maybe this means that `add_status_good_req` and `add_status_bad_req` should be removed? Or maybe we should have a pair of these methods for each kind of status update?

hodgestar:
Idea for better status names:
* `http-requests-to-vumi-go` -> `submitted-to-vumi-go`
* `http-requests-from-vumi-go` -> `received-from-vumi-go`
* `events-from-vumi-go` -> `sent-by-vumi-go`
Thoughts?

vumi/transports/vumi_bridge/vumi_bridge.py
KaitCrawford: a pair of them for each kind of status update at this point seems a bit much to me. I vote for removing them rather

KaitCrawford:
@hodgestar please re-review

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -134,6 +122,9 @@ def handle_outbound_message(self, message):
         if resp.code != http.OK:
             log.warning('Unexpected status code: %s, body: %s' % (
                 resp.code, resp.delivered_body))
+            self.update_status(
+                status='down', component='submitted-to-vumi-go',
+                type='bad_request', message='Bad request sent')
```
hodgestar: Maybe the message here should be `Request rejected` (since we don't know if the request was bad -- only that Vumi Go didn't accept it).

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -142,6 +133,9 @@ def handle_outbound_message(self, message):
         remote_message = json.loads(resp.delivered_body)
         yield self.map_message_id(
             remote_message['message_id'], message['message_id'])
+        self.update_status(
+            status='ok', component='submitted-to-vumi-go',
+            type='good_request', message='Good request sent')
```
hodgestar: Similarly maybe the message here should be `Message accepted by Vumi Go`?

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -323,10 +230,21 @@ def handle_raw_inbound_event(self, request):
             msg = TransportEvent(_process_fields=True, **to_kwargs(data))
             yield self.handle_inbound_event(msg)
             request.finish()
+            if msg.payload["event_type"] == "ack":
+                self.update_status(
+                    status='ok', component='sent-by-vumi-go',
+                    type='vumi_go_sent', message='Sent by VumiGo')
```
hodgestar: I usually spell `Vumi Go` with a space.

KaitCrawford:
@hodgestar updated again

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -323,10 +231,21 @@ def handle_raw_inbound_event(self, request):
             msg = TransportEvent(_process_fields=True, **to_kwargs(data))
             yield self.handle_inbound_event(msg)
             request.finish()
+            if msg.payload["event_type"] == "ack":
+                self.update_status(
+                    status='ok', component='sent-by-vumi-go',
+                    type='vumi_go_sent', message='Sent by Vumi Go')
+            else:
```
hodgestar: Events could also be delivery reports, so we should do something like `elif msg.payload["event_type"] == "nack"` here.

vumi/transports/vumi_bridge/vumi_bridge.py
```diff
@@ -323,10 +231,21 @@ def handle_raw_inbound_event(self, request):
...
+                self.update_status(
+                    status='down', component='sent-by-vumi-go',
+                    type='vumi_go_failed', message='Vumi Go failed to send')
         except Exception as e:
             log.err(e)
             request.setResponseCode(400)
             request.finish()
+            self.update_status(
+                status='down', component='sent-by-vumi-go',
+                type='bad_request', message='Bad request sent')
```
hodgestar: Maybe the type here should be `vumi_go_bad_event` and the message `Bad event received from Vumi Go`.

hodgestar:
Left what I hope are the last two minor comments!

vumi/transports/vumi_bridge/vumi_bridge.py
KaitCrawford: @hodgestar Didn't we decide we weren't going to handle delivery reports at this juncture? Although if this is all that's required I'm happy to add them.
What would a `nack` event_type mean and what should we do when we receive them?

vumi/transports/vumi_bridge/vumi_bridge.py
KaitCrawford: Just a clarification, should events have underscores or dashes or doesn't it matter?

vumi/transports/vumi_bridge/vumi_bridge.py
hodgestar: Events come in three kinds -- acks, nacks and delivery reports. Acks are
"message successfully put in pillar box at post office". Nacks are "message
spat out by pillar box".

Currently the code publishes failed statuses for all events that aren't
acks. It should only do that for nacks.

vumi/transports/vumi_bridge/vumi_bridge.py
hodgestar: Do you mean in the "type" field? If so, it's not vitally important whether
there are underscores or dashes -- we should just be readable and
consistent (we seem to have used underscores for types so far and dashes
for component names, which I think is okay).

KaitCrawford:
@hodgestar ready for another round

hodgestar:
:+1: :cake:

Also, I am honoured to present @KaitCrawford with The Monthly :camel: Trophy for continuing with this PR all the way to the end!

KaitCrawford:
what is a :camel: trophy and should I be scared?

hodgestar:
It is a trophy for successfully crossing the desert. :)

PR #1029, opened 2016-03-22
Fix manhole middleware tests.

hodgestar:
These tests suddenly run under Travis with Twisted 16.0 (yay!), but are very broken (eep!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ install:
   - "pip install 'pip>=7.1.0'"
   - "pip install ${TWISTED_VERSION}"
   # If requested, install PyCrypto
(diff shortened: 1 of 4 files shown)


hodgestar:
This version passes tests locally for me on Twisted 16.0. Will need to see what happens with Twisted 13.2 in Travis.

hodgestar:
The tests are skipped for Twisted 13.2 on Travis as they always were and now pass on Twisted 16.0. Given that the tests were completely broken (on all Twisted versions before), these feels like it's already a major improvement.

Ready for review.

hodgestar:
Tests now run & pass for Twisted 13.2 too.

hodgestar:
@rudigiesler @KaitCrawford Would you mind reviewing?

rudigiesler:
:+1:

PR #1031, opened 2016-04-05
Remove references to redis sub managers

rudigiesler:
For the txredis manager, a list of submanagers is maintained for updating the clients of those submanagers in the case of a reconnection. This can cause problems if many sub managers are used, as when the sub manager goes out of scope, it is still being held on in memory by the list on the base manager.

The plan is to create a client proxy object that each submanager points to, which contains a reference to the client. That way, when we update the client in a reconnection, we only need to update the reference in the client proxy object, and not all the references for each submanager, meaning we no longer need the list of submanagers.


vumi/persist/redis_base.py
```diff
@@ -55,14 +55,24 @@ def __new__(meta, classname, bases, class_dict):
...
+    def __init__(
+            self, client, config, key_prefix, key_separator=None,
+            client_proxy=None):
         if key_separator is None:
             key_separator = ':'
-        self._client = client
+        if client_proxy is None:
+            self._client_proxy = ClientProxy(client)
+        else:
+            self._client_proxy = client_proxy
```
rudigiesler: This ensures that submanagers use the same client, while root managers create a client proxy. I'm not sure of a better way of doing this without changing the API. Ideas?

rudigiesler:
Test failures are unrelated to this PR, and are being resolved in https://github.com/praekelt/vumi/pull/1032

vumi/persist/redis_base.py
hodgestar: This looks fine, I think.

Perhaps we should assert that only one of `client` or `client_proxy` is not `None`? (and then pass in `None` for the client when creating sub-managers?

hodgestar:
Left one minor comment, otherwise code looks good to me.

@rudi Is this successfully preventing us leaking sub managers?

rudigiesler:
@hodgestar Comment addressed.

Yes this does prevent us from leaking sub managers. That list no longer exists, so the sub managers can disappear. I've tested this branch with Junebug benchmarks, and there are no longer any sub managers left over.

hodgestar:
:+1:

PR #1033, opened 2016-04-08
Fix fake_connection abortConnection under Twisted 16.1

hodgestar:
Twisted 16.1 adds its own abortConnection, so we need to improve our wrapping of abortConnection for fake connections.

### DIFF
--- a/vumi/tests/fake_connection.py
+++ b/vumi/tests/fake_connection.py
@@ -220,14 +220,12 @@ def patch_transport_abortConnection(transport, protocol):
     Patch abortConnection() onto the transport if it doesn't already have it.
     (`Agent` assumes its transport has this.)
     """
-    _old_abortConnection = getattr(transport, 'abortConnection', None)
+    _old_abortConnection = getattr(
+        transport, 'abortConnection', transport.loseConnection)
 
     def abortConnection():
         protocol._fake_connection_aborted = True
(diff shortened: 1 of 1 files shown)


vumi/tests/fake_connection.py
```diff
@@ -220,10 +220,16 @@ def patch_transport_abortConnection(transport, protocol):
     Patch abortConnection() onto the transport if it doesn't already have it.
     (`Agent` assumes its transport has this.)
     """
+    _old_abortConnection = getattr(transport, 'abortConnection', None)
```
jerith: `getattr(transport, 'abortConnection', transport.loseConnection)` ?

hodgestar: Nice! Added.

vumi/tests/fake_connection.py
```diff
@@ -220,10 +220,14 @@ def patch_transport_abortConnection(transport, protocol):
     Patch abortConnection() onto the transport if it doesn't already have it.
     (`Agent` assumes its transport has this.)
     """
```
jerith: The docstring should probably explain that even if the method is there, we may need to track that it's an abort so we can swap out the connectionLost reason.

jerith:
:+1: after the docstring update.

hodgestar:
Docstring updated.

PR #1036, opened 2016-04-14
Vumi Bridge transport can't access Vumi Go over HTTPS

hodgestar:
Vumi Go's SSL certificate relies on a weak (SHA1) intermediate certificate. This fails to be verified by a lot of server side software (including Twisted) but is accepted by current browsers. Vumi Go's certificate will be updated in October at the latest, but until then the bridge transport needs to accept weaker certificates.

### DIFF
--- a/vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
+++ b/vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
@@ -37,11 +37,7 @@ class TestGoConversationTransportBase(VumiTestCase):
         transport = yield self.tx_helper.get_transport(defaults, start=False)
         transport.agent_factory = self.fake_http.get_agent
         yield transport.startWorker()
-        yield self.setup_transport(transport)
         returnValue(transport)
-
-    def setup_transport(self, transport):
-        pass
 
     @inlineCallbacks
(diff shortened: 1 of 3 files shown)


hodgestar:
@JayH5 @imsickofmaps Ready for review.

vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
```diff
@@ -178,6 +177,11 @@ def test_receive_bad_event(self):
         self.assertEquals(status['message'], 'Bad event received from Vumi Go')
 
     @inlineCallbacks
+    def test_weak_cacerts_installed(self):
+        yield self.get_configured_transport()
+        self.assertEqual(os.environ["SSL_CERT_FILE"], certifi.old_where())
```
hodgestar: Not sure if there is a clean way to write a better test for this.

hodgestar:
And @jerith ^^^

hodgestar:
@morgs Since this involves SSL, I thought I'd ask you whether you think it's okay to use the weaker CA certificates for now?

jerith:
:+1: although we should test it against some real HTTPS servers before releasing.

PR #1035, opened 2016-04-14
Vumi Bridge transport fails to shut down cleanly if it doesn't start up.

hodgestar:
The teardown method needs to not assume that `.web_resource` is present.

### DIFF
--- a/vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
+++ b/vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
@@ -26,7 +26,7 @@ class TestGoConversationTransportBase(VumiTestCase):
         self.add_cleanup(self.finish_requests)
 
     @inlineCallbacks
-    def get_transport(self, **config):
+    def get_transport(self, start=True, **config):
         defaults = {
             'account_key': 'account-key',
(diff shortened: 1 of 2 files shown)


hodgestar:
Ready for review.

vumi/transports/vumi_bridge/tests/test_vumi_bridge.py
```diff
@@ -36,8 +36,8 @@ def get_transport(self, **config):
         defaults.update(config)
         transport = yield self.tx_helper.get_transport(defaults, start=False)
         transport.agent_factory = self.fake_http.get_agent
-        yield transport.startWorker()
-        yield self.setup_transport(transport)
+        if start:
+            yield transport.startWorker()
```
rgiesler: We're losing the `setup_transport` call here. Is that a bad thing?

rudigiesler:
Just the one question from me. Looks good otherwise.

hodgestar:
The `setup_transport` method is currently empty and I'm removing it as part of #1036 (sorry, should have mentioned that in a comment straight away).

rudigiesler:
Ah okay, I see. :+1:

PR #1038, opened 2016-04-18
Set agent_factory in Vumi Bridge transport.

hodgestar:
Currently the `agent_factory` is not set which means the Vumi Bridge transport is unable to make HTTP requests.


hodgestar:
Ready for review.

KaitCrawford:
:+1:

PR #1040, opened 2016-04-19
Update default base_url for Vumi Bridge transport

hodgestar:
The Vumi Bridge transports still uses the old streaming API by default.


hodgestar:
Ready for review.

rudigiesler:
:+1:

PR #1044, opened 2016-06-14
Some Docker updates

JayH5:
I'm considering moving this to its own repo so that it can be updated separately from Vumi itself and Travis can test the container and such things. Anyway, for now here are a few changes.
* Vumi 0.6.8
* Some tweaks to the entrypoint script
* Made the entrypoint script the `CMD` rather than `ENTRYPOINT` this means that if you want to pass arguments to the script you have to do `CMD ["/app/vumi-entrypoint.sh", "arg1", "arg2"]` but it means it's easier to do things like `docker run vumi bash` and get a shell. I prefer having the entrypoint fixed and updating the command. This will require some small changes to one or two images in `mama-ng-dockerconfig`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,5 +7,4 @@ RUN pip install vumi==$VUMI_VERSION
 COPY ./vumi-entrypoint.sh /app/vumi-entrypoint.sh
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


hodgestar:
Plus one on making a separate repository for the docker container.

hodgestar:
:+1:

PR #1045, opened 2016-06-23
Force SMPP data_coding keys to be ints

imsickofmaps:
This is needed because some dynamic config producers are not able to store keys as ints (like Junebug)

### DIFF
--- a/vumi/transports/smpp/processors/default.py
+++ b/vumi/transports/smpp/processors/default.py
@@ -268,7 +268,8 @@ class DeliverShortMessageProcessor(object):
             9: 'shift_jis',
             10: 'iso2022_jp'
         }
-        self.data_coding_map.update(self.config.data_coding_overrides)
+        self.data_coding_map.update(dict((int(key), value) for (
+            key, value) in self.config.data_coding_overrides.items()))
         self.allow_empty_messages = self.config.allow_empty_messages
 
     def dcs_decode(self, obj, data_coding):


imsickofmaps:
@jerith @smn @rudigiesler @justinvdm can someone review this and release if acceptable? 

Apologies for the @ bomb - I discussed this fix with @hodgestar and it's a known issue discussed with @rudigiesler back in May but the fix on Junebug would be quite complex and this is simple.

hodgestar:
We might want to raise a `ConfigError` with a suitable message rather than raising a `TypeError` or `ValueError`.

smn:
A test would be nice! It's in your fork and so I cannot contribute it easily to your repo.

imsickofmaps:
I wasn't sure how you would do tests for this as you haven't got anything testing config that I could see, without duplicating a large chunk of test setup stuff. I did it in my fork because I don't have access to this repo that I know of. Happy to add you to mine if it's quicker to show rather than tell on the adding tests issue?

smn:
@imsickofmaps I would expect the following test to fail for you:

```python
from twisted.internet.task import Clock

from vumi.errors import ConfigError
from vumi.tests.helpers import VumiTestCase
from vumi.transports.tests.helpers import TransportHelper
from vumi.transports.smpp.smpp_transport import SmppTransceiverTransport
from vumi.transports.smpp.tests.fake_smsc import FakeSMSC


class DefaultProcessorTestCase(VumiTestCase):

    def setUp(self):
        self.clock = Clock()
        self.fake_smsc = FakeSMSC()
        self.tx_helper = self.add_helper(
            TransportHelper(SmppTransceiverTransport))

    def test_data_coding_strings(self):
        cfg = {
            'system_id': 'foo',
            'password': 'bar',
            'transport_name': self.tx_helper.transport_name,
            'twisted_endpoint': self.fake_smsc.endpoint,
            'deliver_short_message_processor_config': {
                'data_coding_overrides': {
                    'not-an-int': 'utf-8',
                }
            },
        }
        return self.assertFailure(
            self.tx_helper.get_transport(cfg), ConfigError)
```

PR #1047, opened 2016-07-11
Feature/issue 1046 fake redis expire response error with non int values

wimkeir:


vumi/persist/fake_redis.py
```diff
@@ -502,6 +502,9 @@ def ltrim(self, key, start, stop):
 
     @maybe_async
     def expire(self, key, seconds):
+        # raise ResponseError for non-int values to mimic Redis behaviour
+        if type(seconds) is not int:
```
hodgestar: You need to also allow it to be type `long` (Python has two kinds of ints) and it might be better to do `if isinstance(seconds, (int, long))` to allow sub-classes of these types to also be allowed (it depends a bit what the real redis libraries do).

hodgestar:
Left a minor comment. We also need to fix all the tests that now fail though. :(

PR #1048, opened 2016-07-25
Add support for reading to_addr for destination_number to 6D SMPP processor

hodgestar:
Currently the 6D SMPP USSD message processor reads the USSD to_addr (i.e. the star code) from the short_message PDU field of the PSSR message. We have encountered some instances that instead sent the to_addr in the destination_number PDU field, and we need a configuration option from reading from that field instead.


justinvdm:
:+1: if tests are passing locally, though maybe we should wait for github.com/praekelt/vumi/pull/1049 so tests are passing again after the new twisted release.

hodgestar:
I've confirmed that the Travis failures are all covered by #1049, so landing this.

PR #1049, opened 2016-07-26
Fix tests for Twisted 16.3.

hodgestar:
Twisted 16.3 changed how HTTPChannels work which broken a bunch of tests:
```
vumi.transports.mxit.tests.test_mxit.TestMxitTransport.test_get_request_content_when_missing
===============================================================================
[ERROR]
Traceback (most recent call last):
  File "/home/travis/build/praekelt/vumi/vumi/transports/mxit/tests/test_mxit.py", line 103, in test_get_request_data
    req = Request(None, True)
  File "/home/travis/env-pypy-4.0.1/site-packages/twisted/web/http.py", line 612, in __init__
    self.transport = self.channel.transport
exceptions.AttributeError: 'NoneType' object has no attribute 'transport'
```

PR #1050, opened 2016-07-26
Only use decimal digits in MTN Nigeria XML over TCP transport session IDs.

hodgestar:
Currently we use hexadecimal digits, but non-decimal digits are not allowed.

### DIFF
--- a/vumi/transports/mtn_nigeria/xml_over_tcp.py
+++ b/vumi/transports/mtn_nigeria/xml_over_tcp.py
@@ -1,6 +1,5 @@
-import uuid
+import random
 import struct
-from random import randint
 
 from twisted.web import microdom
 from twisted.internet import reactor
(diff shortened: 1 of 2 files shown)


vumi/transports/mtn_nigeria/xml_over_tcp.py
```diff
@@ -407,17 +407,16 @@ def gen_session_id(cls):
         """
         Generates session id. Used for packets needing a dummy session id.
         """
-        # NOTE: Slicing the generated uuid is probably a bad idea, and will
-        # affect collision resistence, but I can't think of a simpler way to
-        # generate a unique 16 char alphanumeric.
```
justinvdm: */me shakes head at past self*

justinvdm:
:+1: when travis is happy

hodgestar:
I've confirmed that all the Travis errors are covered by #1049 and that the tests pass locally, so landing this.

PR #1052, opened 2016-07-27
Dmark transport should set content to null at the start of a USSD session.

hodgestar:
Currently it passes through the USSD request string directly from Dmark, which confuses applications because they try to treat it as user input.


hodgestar:
Ready for review.

imsickofmaps:
👍  when Travis happy

rudigiesler:
:+1:

PR #1055, opened 2016-08-12
Do not log a warning when SMPP disconnects.

smn:


smn:
@rudigiesler & @justinvdm ready for review. Reduces log spam on Sentry a lot for when we're being disconnected or throttled.

justinvdm:
:+1:

PR #1058, opened 2016-10-25
Replace http_request_full internals with treq

smn:


KaitCrawford:
@smn I just changed the order of things a little

smn:
💥 amazing! Thank you :) :cake: 🍰 :cake: 👍

vumi/utils.py
```diff
@@ -13,12 +13,14 @@
...
 from twisted.internet.defer import succeed
 from twisted.python.failure import Failure
-from twisted.web.client import Agent, ResponseDone, WebClientContextFactory
+from twisted.web.client import Agent, ResponseDone
 from twisted.web.server import Site
 from twisted.web.http_headers import Headers
 from twisted.web.iweb import IBodyProducer
 from twisted.web.http import PotentialDataLoss
 from twisted.web.resource import Resource
+from treq._utils import default_pool, default_reactor
```
JayH5: I don't think it's a good idea to import from a private module.. These methods could change.

JayH5: And you don't seem to be using `default_reactor`.

erikh360:
👍

vumi/utils.py
```diff
@@ -119,13 +121,66 @@ def connectionLost(self, reason):
...
+    client = HTTPClient(agent)
+
+    def handle_response(response):
+        return SimplishReceiver(response, data_limit).deferred
+
+    d = client.request(method, url, headers=headers, data=data)
+    d.addCallback(handle_response)
+
+    if timeout is not None:
+        cancelling_on_timeout = [False]
```
JayH5: Reason that this is a list?

vumi/utils.py
KaitCrawford: when calling `default_pool` we were sending `pool=None` and `persistence=False`. In that case the method will create a new HTTPSConnectionPool. so that is what I did instead.
https://github.com/twisted/treq/blob/master/src/treq/_utils.py#L38-L42

vumi/utils.py
KaitCrawford: @smn perhaps you can answer this?

smn: the why escapes me, this was many years ago, I got it from the old implementation: https://github.com/praekelt/vumi/pull/1058/files/a22e08f7250ab6570fd96b8469d96090696b4560#diff-832ae36feb20f811191dde6be9132638R195

perhaps @jerith remembers?

rudigiesler: Is this not something to do with that `False` in python is a simple variable, where as `[False]` is an object?

```>>> False is False
True
>>> [False] is [False]
False```

**Comment by KaitCrawford** [comment]
@JayH5 ^^ this commit removes usage of `WebClientContextFactory()` from the sandbox and it doesn't appear anywhere else in the repo. I'm not sure if that's enough though?

**Comment by JayH5** [comment]
@KaitCrawford I actually read the `sandbox.py` file.. it says it's deprecated and people should use `vxsandbox` rather which is what we do for Seed. So I'd say maybe actually just don't touch `sandbox.py` in this PR.

File: vumi/utils.py (hunk 2 above)
Comment (jerith) [inline vumi/utils.py]: It's a hack to get a mutable value that can be modified elsewhere:
```
>>> foo = [1]; print foo, id(foo)
[1] 4359988792
>>> foo[0] = 2; print foo, id(foo)
[2] 4359988792
```
This lets us pass `cancelling_on_timeout` through a bunch of other code (which can modify the contents) and then check it later to see if it's been changed. Note that this pattern also requires modifications to be of the form `foo[0] = 3` (update the existing list) rather than `foo = [3]` (replace the list with a new one).

jerith: In this particular case, we use `cancelling_on_timeout` to get information from `cancel_on_timeout()` (which is scheduled to cancel the request after some time) into the `raise_timeout()` errback so we can raise a timeout exception instead of a generic cancellation exception or whatever.

JayH5:
Looks like this is definitely fixed in `vxsandbox`, even if treq isn't used there: https://github.com/praekelt/vumi-sandbox/blob/develop/vxsandbox/resources/http.py#L23-L41

PR #1062, opened 2016-12-14
Cleanup the requests in DMark USSD if we've timed out

smn:
My mailbox has a few thousand reasons why this is a good idea.


rudigiesler (approved):

justinvdm:
👍  but wondering if we want to at least log something (maybe at a log level that sentry doesn't care about)

justinvdm:
:+1: when travis is happy

PR #1064, opened 2017-01-09
Fix message sender for new riak client

rudigiesler:
With the new riak client, we will have to change the way that we do cleanup for tests. We need to create a new manager to do the cleanup, instead of using the existing closed manager.

### DIFF
--- a/vumi/tests/helpers.py
+++ b/vumi/tests/helpers.py
@@ -1454,7 +1454,12 @@ class PersistenceHelper(object):
                 unclosed_managers.append(manager)
             if purge:
                 try:
-                    yield self._purge_riak(manager)
+                    # Create a new client for cleanup
+                    purge_manager = self.get_riak_manager(config={
+                        'bucket_prefix': manager.bucket_prefix
+                    })
+                    yield self._purge_riak(purge_manager)
+                    yield purge_manager.close_manager()
                 except ConnectionRefusedError:
                     pass
             yield manager.close_manager()
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
         'riak>=2.1',
         'txJSON-RPC==0.3.1',
         'txTwitter>=0.1.4a',
-        'treq',
+        'treq<16.12.0',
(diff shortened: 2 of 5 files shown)


vumi/tests/helpers.py
```diff
@@ -1454,7 +1454,12 @@ def cleanup(self):
                 unclosed_managers.append(manager)
             if purge:
                 try:
-                    yield self._purge_riak(manager)
+                    # Create a new client for cleanup
+                    purge_manager = self.get_riak_manager(config={
+                        'bucket_prefix': manager.bucket_prefix
+                    })
```
rudigiesler: This is the part I'm not too certain about. Is there anything other than the bucket prefix that we need to worry about for cleaning up for tests?

justinvdm:
:+1: when travis is happy

Seems unlikely, but do you know whether the change to the new riak client affects our non-test code? I'm guessing it would be the case if anywhere in our code we tried to purge data for a closed manager.

rudigiesler:
Tests are failing, I'm guessing that new treq doesn't like old twisted

rudigiesler:
@justinvdm Thanks. I'm assuming that we don't purge data in non-test code, so that it wouldn't be a problem.

vumi/tests/helpers.py
justinvdm: Hmm, I have no idea, sorry for missing this. @jerith would probably know

jerith: Not unless we're talking to multiple Riak clusters something. In that case, we'd need more extensive changes to the test infrastructure anyway.

jerith:
Anything that uses a closed manager in non-test-cleanup code should fail its tests with the new behaviour. There might be a bug or two where this happens in code paths we don't test, but that's true of any code using things we might update.

rudigiesler:
@justinvdm I've fixed the treq problem by pinning it to an older version that still supports twisted 13.2.0. Can you give it another look?

justinvdm:
👍 !

PR #1069, opened 2017-03-31
Fix address for DMark USSD if asked

smn:


vumi/transports/dmark/dmark_ussd.py
```diff
@@ -19,6 +19,10 @@ class DmarkUssdTransportConfig(HttpRpcTransport.CONFIG_CLASS):
         " expires.",
         default=600, static=True)
 
+    fix_to_addr = ConfigBool(
+        "Whether or not to ensure that the to_addr is always starting with a "
+        "* and ending with a #", default=False, static=True)
```
jerith: I don't think this needs to be static. IIRC, non-static options can be configured statically or dynamically.

smn: it doesn't change per message received, it's a config thing on their side so I would think it is static? That way I can use the `get_static_config()` helper and it's still cheap to do so.

jerith: Static configs are only required for things that are used in transport setup or whatever. Dynamic configs are looked up dynamically in places where dynamic overrides are available (such as vumi-go channel definitions) but fall back to a static lookup otherwise. The per-message overhead is negligible.

smn: Inclined to keep as is as all the config vars in there are currently marked as static and I'm using the `get_static_config()` convenience helper.

PR #1070, opened 2017-05-11
Make multipart reference rollover configurable

rudigiesler:
Currently, we rollover the `sar_msg_ref_num` at `0xFFFF`. We want to make this value configurable, for cases where, for example, AAT want us to roll it over at `0xFF` to not affect character space.

### DIFF
--- a/vumi/transports/smpp/processors/default.py
+++ b/vumi/transports/smpp/processors/default.py
@@ -470,6 +470,10 @@ class SubmitShortMessageProcessorConfig(Config):
         "If `True`, messages longer than 140 bytes will be sent as a series "
         "of smaller messages with the user data headers. Default is `False`.",
         default=False, static=True)
+    multipart_sar_reference_rollover = ConfigInt(
+        "The maximum value to set for the reference number of a multi part "
+        "SMS. When this value is reached, the number will rollover to 0.",
+        default=0xFFFF, static=True)
 
     def post_validate(self):
         long_message_params = (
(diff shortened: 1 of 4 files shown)


vumi/transports/smpp/processors/default.py
```diff
@@ -470,6 +470,10 @@ class SubmitShortMessageProcessorConfig(Config):
         "If `True`, messages longer than 140 bytes will be sent as a series "
         "of smaller messages with the user data headers. Default is `False`.",
         default=False, static=True)
+    multipart_sar_reference_rollover = ConfigInt(
+        "The maximum value to set for the reference number of a multi part "
+        "SMS. When this value is reached, the number will rollover to 0.",
+        default=0xFFFF, static=True)
```
jerith: This is actually a bitmask rather than a maximum value.

Maybe make the default `0x10000` and document that the reference resets to zero when it reaches that value or something? Then we can go back to using `%` and it will behave as expected for all possible config values.

Alternatively, maybe this should be `multipart_sar_reference_octets` with `1` and `2` as possible values? Then we can check that and either `& 0xFF` or `% 0xFFFF` as appropriate.

rudigiesler: Is a bitmask and a maximum value not the same? If I set a maximum of `255`, I would expect it to go  0 -> 255, and then rollover. I wouldn't expect it to only go to 254.

jerith: They're only the same for values of the form (2**n)-1. For example: `[i & 0xFE for i in range(260)]` -> `[0, 0, 2, 2, ..., 252, 252, 254, 254, 0, 0, 2, 2]`

rudigiesler: Ah right, of course. Okay let me fix that.

rudigiesler:
@jerith Added a test for that. I looked at the other processors, and they all inherit from default, and they all call the `send_short_message` function, and none of them overwrite that function.

PR #1072, opened 2017-05-29
Add session_id to helper_metadata for dmark

smn:
Some external systems, like RapidPro, rely on the channels to provide a 
reliable session identifier. For DMark we'll be relaying whatever transaction
id the aggregator gives us.


smn:
@rudigiesler ready for review

PR #1073, opened 2017-07-12
Minor changes to make sure Vumi installs on python 3

smn:
This doesn't give any guarantees on it being able to run on python 3 though.


smn:
@KaitCrawford ready for review

PR #1075, opened 2017-09-11
Improve Python 3 support

alexmuller:
This is probably really boring to review but hopefully it's fairly uncontroversial. Every change here acts identically between between Python 2.7 and Python 3.6.

I've tried to group the changes into logical commits and have added more details in each commit message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ python:
   - "3.6"
 node_js:
   - "0.10"
+env:
+  - RUN_TESTS=true
 matrix:
   include:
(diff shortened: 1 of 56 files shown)


JayH5:
LGTM so far.

A couple of minor things:
* We should probably test against Python 3.5 as well since that's what's in Debian Jessie/Ubuntu Xenial.
* I would probably do `from __future__ import print_function` where we print.

alexmuller:
Thanks! Pushed two more commits to address your comments.

PR #1078, opened 2020-04-30
Add turn transport

erikh360:


vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+                    loc = message["location"]
+                    content = 'geo:{},{}'.format(loc['latitude'], loc['longitude'])
+
+                yield self.publish_inbound_func(
+                    transport_name=self.transport_name,
+                    transport_type='sms',
+                    message_id=message_id,
+                    transport_metadata={
+                        'original_message_id': message_id,
+                        'timestamp': iso8601(message["timestamp"]),
```
rudigiesler, line 70: Vumi messages do have a timestamp, I'm wondering if we should put the timestamp on the message rather than the metadata

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+                        'network_id': request.args['provider'][0],
+                        'keyword': request.args['keyword'][0],
```
rudigiesler, line 72: I'm wondering if instead of requiring these, we just convert any querystring arguments to a dict, and add it to the transport metadata.

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+def iso8601(turn_timestamp):
+    # TODO: this will be a unix timestamp
+    if turn_timestamp:
+        ts = datetime.strptime(turn_timestamp, '%Y.%m.%d %H:%M:%S')
+        return ts.isoformat()
+    else:
+        return ''
+
+
+def normalize_outbound_msisdn(msisdn):
```
rudigiesler, line 33: Maybe a better name would be `format_msisdn_for_whatsapp` or something like that, since it takes the MSISDN, and formats it in the specific format for whatsapp.

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+    if msisdn.startswith('+'):
+        return msisdn.replace('+', '')
+    else:
+        return msisdn
```
rudigiesler, line 37: We can probably simplify this as `return msisdn.lstrip("+")`

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+
+    @inlineCallbacks
+    def do_render(self, request):
+        request.setResponseCode(http.CREATED)
+        # request.setHeader('Content-Type', 'text/plain')
+        try:
+            # TODO: validate HMAC secret
+
+            for message in request.args.get("messages", []):
+                message_id = '%s.%s' % (self.transport_name, message["id"])
```
rudigiesler, line 56: Are whatsapp message IDs not guaranteed to be unique across numbers? Or are we just being overly cautious?

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+                content = ''
+                if message['type'] == 'text':
+                    content = message["text"]["body"]
+                elif message['type'] == 'location':
+                    loc = message["location"]
+                    content = 'geo:{},{}'.format(loc['latitude'], loc['longitude'])
```
rudigiesler, line 62: I think in this case, we rather want to set the content to `None`, and put the location data in the metadata.

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+                    delivery_status = 'pending'
+                elif event["status"] == 'failed':
+                    delivery_status = 'failed'
+                elif status == 'delivered':
+                    delivery_status = 'delivered'
+                else:
+                    continue
+
+                yield self.publish_event_func(
+                    user_message_id=message_id,
```
rudigiesler, line 93: This should be the ID of the message that this event is for, not the ID of the event.

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+
+            for event in request.args.get("statuses", []):
+                message_id = '%s.%s' % (self.transport_name, event['id'])
+
+                if event["status"] == 'sent':
+                    delivery_status = 'pending'
+                elif event["status"] == 'failed':
+                    delivery_status = 'failed'
+                elif status == 'delivered':
+                    delivery_status = 'delivered'
```
rudigiesler, line 88: We can probably add `read` as a `delivered` status.

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+        try:
+            response = yield http_request_full(
+                self.config['url'], message_params.to_json(), {
+                    'User-Agent': ['Vumi Turn Transport'],
+                    'Content-Type': ['application/json'],
+                    'Authorization': ['Bearer {}'.format(TURN_TOKEN)],
+                    }, 'POST', agent_class=self.agent_factory)
+        except ConnectionRefusedError:
+            log.msg("Connection failed sending message:", message)
+            raise TemporaryFailure('connection refused')
```
rudigiesler, line 187: Instead of using this old vumi method (it has it's issues), I think we should rather rely on `treq`

vumi/transports/turn/turn.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class TurnTransport(Transport):
+
+    agent_factory = None  # For swapping out the Agent we use in tests.
+
+    def mkres(self, cls):
+        resource = cls(self.config, self.publish_delivery_report, self.publish_message)
+        self._resources.append(resource)
+        return (resource, self.config['web_%s_path' % (path_key,)])
```
rudigiesler, line 141: If we use the `HttpRpcTransport` as a base for this (even though it's not RPC), then we don't need to handle all of this ourselves. You can look at https://github.com/praekeltfoundation/vumi-vas2nets/blob/develop/vxvas2nets/vas2nets_sms.py for a good example of this.

